PR #3, opened 2013-08-23
Handle exception create index request

yrjsg:
Below is the situation when exception happened, flume cannot commit txn and keep rolling back.

screenshot from 2013-08-23 12 18 32


harishreedharan:

szaboferee:
@frankyaorenjie Thanks again for your contribution! Since the source repository/branch seems to be deleted, do you think that this pull request can be closed? If it is, could you please close it?

yrjsg:
@szaboferee let's close it, since it has been too long time. And my origin code has already merged into flume.

PR #4, opened 2013-12-24
Add methods to retrieve information about the file

PieterJanVanAeken:
The getFileName() and getFilePath() methods are useful for a custom  deserializer that works with a ResettableFileInputStream. At the moment, there is no way in a Deserializer to find out which file it is processing. The other option is to make the FileInfo available in the Deserializer but imo this looks like the better of the two options.


harishreedharan:

szaboferee:
@PieterJanVanAeken  Thanks again for your contribution! Since the source repository/branch seems to be deleted, do you think that this pull request can be closed? If it is, could you please close it?

asfgit:

szaboferee:
No Jira ticket and source branch is deleted. I am closing this PR. Feel free to reopen if needed.

PR #6, opened 2014-09-29
Add failover log4j appender

EdwardsBean:
Add a log4j appender base on flume FailoverRpcClient. Provide two parameters:MaxAttempts and Hosts.


harishreedharan:

szaboferee:
@EdwardsBean  Thanks again for your contribution! Since the source repository/branch seems to be deleted, do you think that this pull request can be closed? If it is, could you please close it?

PR #7, opened 2014-10-17
change FLUME_HOME to fix the LIBRARY NOT FOUND bug

shadyxu:
Assigning `FLUME_HOME` to `$(cd $(dirname $0)/..; pwd)` will cause an `org.apache.flume.node.Application not found` error. This can be fixed by elimilated the `;pwd` path.


harishreedharan:

PR #10, opened 2014-12-13
Changes to accept directory which stores config file as one of the param

bijugs:
Fix for [FLUME-2573](https://issues.apache.org/jira/browse/FLUME-2573)


asfgit:

PR #11, opened 2015-02-11
added ProxySupport for TwitterSource

schmiegelow:
This is a fix proposal for the following issue
https://issues.apache.org/jira/browse/FLUME-2618


asfgit:

PR #12, opened 2015-02-13
Fix tests when default locale is not english

sziep:
When running the tests on a system with a non english / us locale, some tests fail.
An explicit Locale is used now for the respective tests.
Readme was updated to note that an increased ulimit is needed for the tests to run.


harishreedharan:

asfgit:

PR #16, opened 2015-04-01
fix: netcat source tcp connection is not closed

ericx10ng:
when client using tcp short connection, netcat source tcp connections were not closed. flume had lots of tcp close_wait until  process exited!


harishreedharan:

PR #17, opened 2015-04-17
Update EventHelper.java

martinschaef:
Avoid null pointer exception in line 71 for the case "if (event == null || event.getBody() == null)" in line 44/45


harishreedharan:
Flume does not accept PRs. Could you file a jira here: https://issues.apache.org/jira/browse/FLUME and submit this as a patch? Thanks!

martinschaef:
I'm not familiar with jira ... I guess I created an issue and not a patch. But its a one-liner anyway.

PR #19, opened 2015-06-01
Fix Configuration failed bug.

chenzhiguo:
Fix Configuration failed bug.


harishreedharan:

asfgit:

rgoers:
The elasticsearch sink has been removed from the core Kafka repo as it was seriously out of date and its license is problematic. Any enhancements should be directed to the flume-search repo.

PR #21, opened 2015-07-03
Add a new ElasticSearchLogStashEventSerializer 

omallassi:
to handle kibana v12 version of model + be sure to not send all message properties as String but also as long / double if possible (impact the json format)


harishreedharan:

omallassi:
@harishreedharan done there https://issues.apache.org/jira/browse/FLUME-2784

PR #23, opened 2015-08-05
Update KafkaSource to reads multiple topics

guitoun3:
This change allows to read from multiple topics by using a TopicFilter with a WhiteList.


harishreedharan:

guitoun3:
Ok, Jira is there : https://issues.apache.org/jira/browse/FLUME-2755

PR #24, opened 2015-09-01
FLUME-2781 proposed implementation

gherreros:
Implemented a tiny change so "put" in the channel is consistent with "take". If parseAsFlumeEvent is false it uses the event body text instead of an Avro Flume event object.


harishreedharan:
We don't do pull requests. Could you please create a patch file and attach it to the jira you created? Thanks!

gherreros:
Added patch file to the ticket.

alvarogala:
Hi gherreros, I have been fighting this trouble for some days. Thank you for the patch!

asfgit:

rgoers:
It appears that this has already been applied, or some variation of it.

PR #25, opened 2015-11-06
check message size before send to kafka

darionyaphet:
Kafka Sink fetch message from channel and use Kafka Producer to send message to Kafka Broker . When Kafka Broker's config `message.max.bytes` and message is too long Flume Kafka Sink will failed to send message. 

I add a check before message send to Kafka Broker . The too long message will be discard and record log with message size.
You also can config the messageMaxSize in properties file.

I Create a JIRA Ticket:https://issues.apache.org/jira/browse/FLUME-2837

thanks


harishreedharan:
Unfortunately we do not accept Pull Requests for flume. Could you submit this as a patch on the jira, please? Thanks!

darionyaphet:
hi @harishreedharan  I had submit this patch at jira .  thank you ~

asfgit:

PR #29, opened 2016-01-03
Remove capacity parameters from Kafka Channel example config

peterableda:


peterableda:
I will create a patch and attach it to a JIRA

PR #32, opened 2016-01-19
Size based rolling support for file sink

praste:
This fixes #FLUME-2856.
Also added support use a fixed naming pattern for files.
Support for allowing to delete old sink files is also added.

Updated documentation for above mentioned changes.

 To support `maxHistory` feature (i.e. to delete older files) I had to be able to predict file names. I introduced new required property for `RollingFileSink` called `fileName`. This is the only backwards incompatible change from end users point of view. I have also made a couple of backwards incompatible changes to `PathManager` class. However `PathManager` class is not being used by any class other than `RollingFileSink` class.

I am open to rewrite some of the changes if people think refactoring will cause any issues.

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/formatter/output/PathManager.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/formatter/output/PathManager.java
@@ -24,33 +24,31 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 public class PathManager {
 
-  private long seriesTimestamp;
   private File baseDirectory;
   private AtomicInteger fileIndex;
-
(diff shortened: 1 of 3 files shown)


praste:
BTW I don't know if I was supposed to send a pull request of attach patch to the JIRA. 
It's been more than a month since I have attached patch to the  [FLUME-2856](https://issues.apache.org/jira/browse/FLUME-2856). I have also sent a couple of message to flume-dev list and tried sending emails to flume a couple of moderators/contributors email addressed documented on the flume's team page, with no avail.

praste:
any advise????

asfgit:

PR #38, opened 2016-02-24
Flume 2866

phlantin:
This resolves FLUME-2866.

This adds a property "fileTimeMinOffsetSeconds" in the Spooling Directory Source. This prevents processing of files that are not older or newer than the amount of seconds defined by this property. The default value is "0" seconds, which preserves current behavior.


asfgit:

PR #39, opened 2016-02-24
FLUME-2459: Spooling Directory Source support for compressed files

ro7m:
Flume has a fantastic source for spooling files, however, many systems store relevant files as compressed files. This change is to enable flume Spooling Directory Source to support GZip  compressed files just as if they were plain text (e.g. read line by line).


harishreedharan:
Could you please submit this as patch against the trunk branch on the jira? We don't do PRs yet on Flume

PR #41, opened 2016-02-25
FLUME-2459 Spooling Directory Source support for gzip files

ro7m:
Flume has a fantastic source for spooling files, however, many systems store relevant files as compressed files.This change enables flume spooling directory Source to read gzip (.gz) compressed files just as if they were plain text (e.g. read line by line).


styg:
I had made a simple implementation, deal directly with gz data source:
https://github.com/styg/bumblebee-ETL/tree/master/src/main/java/com/gsta/bigdata/etl/flume

asfgit:

PR #43, opened 2016-03-11
modify rotate way.

skadilover:
customer can rotate file as :


asfgit:

PR #45, opened 2016-03-15
Fixes to address 2K message size limit when using SyslogUDPSource.

ryanpersaud:
Please see https://issues.apache.org/jira/browse/FLUME-2885.  Currently, the SyslogUDPSource truncates messages > 2K.  With this fix, the default behavior remains the same (2K is used for the buffer size), but if a user specifies a value for datagramSize in the flume configuration, then that size is used instead.  If we want to use AdaptiveReceiveBufferSizePredictor to avoid allocating buffers that are larger than necessary, 'someone' is going to have to develop a more comprehensive fix.  This at least allows for udp syslog users with messages larger than 2K to process their messages without truncation.

I have deployed this fix in two environments where syslog messages can be larger than 2K, and it works as expected.

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/SyslogSourceConfigurationConstants.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/SyslogSourceConfigurationConstants.java
@@ -51,6 +51,8 @@ public final class SyslogSourceConfigurationConstants {
 
   public static final String CONFIG_BATCHSIZE = "batchSize";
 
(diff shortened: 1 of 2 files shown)


ryanpersaud:
Hmm, it occurs to me that I should probably set datagramSize to the min of datagramSize and maxsize.

asfgit:

maralla:
Is this project still alive?

szaboferee:
Yes

PR #52, opened 2016-06-09
FLUME-2922 Sync SequenceFile.Writer before calling hflush

kevinconaway:
@harishreedharan will you please review?

### DIFF
--- a/flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
@@ -1366,6 +1366,8 @@ public class TestHDFSEventSink {
     txn.close();
 
     sink.process();
+    // Sink is _not_ closed.  The file should remain open but
+    // the data written should be visible to readers via sync + hflush
     FileStatus[] dirStat = fs.listStatus(dirPath);
     Path[] paths = FileUtil.stat2Paths(dirStat);
 
--- a/flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
@@ -1323,6 +1323,64 @@ public class TestHDFSEventSink {
     fs.close();
   }
 
+  @Test
+  public void testBlockCompressSequenceFileWriterSync() throws IOException, EventDeliveryException, InterruptedException {
+    String hdfsPath = testPath + "/sequenceFileWriterSync";
+    String body = "test event-" + System.nanoTime();
+
+    Configuration conf = new Configuration();
+    FileSystem fs = FileSystem.get(conf);
+
+    // Since we are reading a partial file we don't want to use checksums
+    fs.setVerifyChecksum(false);
+    fs.setWriteChecksum(false);
+
+    Path dirPath = new Path(hdfsPath);
+    fs.delete(dirPath, true);
+    fs.mkdirs(dirPath);
+
+    Context context = new Context();
+    context.put("hdfs.path", hdfsPath);
+    // Ensure the file isn't closed and rolled, we are only going to send one event
+    context.put("hdfs.rollCount", "10");
+    context.put("hdfs.rollSize", "0");
+    context.put("hdfs.rollInterval", "0");
+    context.put("hdfs.batchSize", "1");
+    context.put("hdfs.fileType", "SequenceFile");
+    context.put("hdfs.codeC", "BZip2Codec"); // Compression codec that doesn't require native hadoop libraries
+    context.put("hdfs.writeFormat", "Writable");
+    Configurables.configure(sink, context);
+
+    Channel channel = new MemoryChannel();
+    Configurables.configure(channel, context);
+
+    sink.setChannel(channel);
+    sink.start();
+
+    Transaction txn = channel.getTransaction();
+    txn.begin();
+    Event event = new SimpleEvent();
+    event.setBody(body.getBytes());
+    channel.put(event);
+    txn.commit();
+    txn.close();
+
+    sink.process();
+    FileStatus[] dirStat = fs.listStatus(dirPath);
+    Path[] paths = FileUtil.stat2Paths(dirStat);
+
+    Assert.assertEquals(1, paths.length);
+
+    SequenceFile.Reader reader = new SequenceFile.Reader(conf, SequenceFile.Reader.stream(fs.open(paths[0])));
+    LongWritable key = new LongWritable();
+    BytesWritable value = new BytesWritable();
+
+    Assert.assertTrue(reader.next(key, value));
+    Assert.assertArrayEquals(body.getBytes(), value.copyBytes());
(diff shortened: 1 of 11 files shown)


harishreedharan:
@kevinconaway Can you please submit this as a patch on the jira? We don't do PRs yet.

kevinconaway:
@harishreedharan Is it sufficient to grab the patch from this PR?  You can add `.patch` to the URL to get the patchfile:

https://github.com/apache/flume/pull/52.patch

kevinconaway:
Nm, I just uploaded the patch to Jira.  Thanks!

harishreedharan:
I am not entirely sure what the ASF policy these days is. Earlier, unless it was attached to a jira the copyright was still with the author, so it had to be a patch on the jira (but with the new github integration, that might have changed - I am not sure, so I will commit the patch from the jira later today).

kevinconaway:
@jarcec is this something that you're able to review?

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1323,6 +1323,64 @@ public void testCloseOnIdle() throws IOException, EventDeliveryException, Interr
     fs.close();
   }
 
+  @Test
```
mpercy, line 1334: Please add a comment at the top of the test with a description. You can leave out @throws clauses and all of that junk that is not really additionally helpful to developers reading the test code

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1323,6 +1323,64 @@ public void testCloseOnIdle() throws IOException, EventDeliveryException, Interr
...
+    fs.delete(dirPath, true);
+    fs.mkdirs(dirPath);
+
+    Context context = new Context();
+    context.put("hdfs.path", hdfsPath);
+    // Ensure the file isn't closed and rolled, we are only going to send one event
+    context.put("hdfs.rollCount", "10");
+    context.put("hdfs.rollSize", "0");
+    context.put("hdfs.rollInterval", "0");
+    context.put("hdfs.batchSize", "1");
```
mpercy, line 1372: Please increase your batch size to at least 2 events

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1323,6 +1323,64 @@ public void testCloseOnIdle() throws IOException, EventDeliveryException, Interr
...
+    Configurables.configure(channel, context);
+
+    sink.setChannel(channel);
+    sink.start();
+
+    Transaction txn = channel.getTransaction();
+    txn.begin();
+    Event event = new SimpleEvent();
+    event.setBody(body.getBytes());
+    channel.put(event);
```
mpercy: The test is pretty good but please make the following modifications:
1. Add multiple events to the channel.
2. Call process() multiple times.
3. Ensure that the file is readable by a reader after multiple batches have been flushed.
4. Test all of the following codecs: bzip2, gzip, snappy

You probably would want to extract a helper method to avoid copy / paste for testing all the codecs

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1323,6 +1323,64 @@ public void testCloseOnIdle() throws IOException, EventDeliveryException, Interr
     fs.close();
   }
 
+  @Test
+  public void testBlockCompressSequenceFileWriterSync() throws IOException, EventDeliveryException, InterruptedException {
```
mpercy: Please also rebase on trunk. We've added a checkstyle plugin to the Maven build that will run and tell you what's wrong. In this case, the line exceeds 100 chars.

A reasonable way for you to build Flume and run your test while invoking checkstyle is:

mvn clean install -DfailIfNoTests=false -Dtest=TestHDFSEventSink

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
kevinconaway: > Test all of the following codecs: bzip2, gzip, snappy

Snappy and gzip require the native hadoop libraries.  Do you know of a work around here to enable that?

kevinconaway: > Add multiple events to the channel.
> Please increase your batch size to at least 2 events

The idea behind this test (and this is why your request for documentation is a good one ;)) is to simulate what happens when a batch of events is written to the file (and thus hsync'd to hdfs) but the file is _not_ yet closed.  In that case, the data that we wrote should still be readable.  With that in mind, I thought that writing one event (and thus a corresponding batch size of one) would suffice.  Is there a scenario that I'm missing where multiple events would be different?

Thanks!

mpercy: > > Test all of the following codecs: bzip2, gzip, snappy
> 
> Snappy and gzip require the native hadoop libraries. Do you know of a work around here to enable that?

I believe there are pure Java versions of some compression strategies, particularly deflate and gzip. Maybe not snappy though... Please try deflate and gzip and let me know if it doesn't appear to work.

>  I thought that writing one event (and thus a corresponding batch size of one) would suffice. Is there a scenario that I'm missing where multiple events would be different?

I want to make sure that multiple batches concatenated is still readable in all cases. I suppose more than one event per batch doesn't add much coverage, but certainly we need to test multiple batches in order to ensure that the SequenceFile framing is working properly across multiple sync() calls on the same file.

kevinconaway: > I believe there are pure Java versions of some compression strategies, particularly deflate and gzip. Maybe not snappy though... Please try deflate and gzip and let me know if it doesn't appear to work

Deflate works without the native libraries.  Snappy and gzip do not.  The gzip check is actually done by the `SequenceFile.Writer` for some reason: https://github.com/apache/hadoop/blob/bad85f3e3a53b5fb692c77df6c11dcf968ad7e62/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/SequenceFile.java#L1151

> I want to make sure that multiple batches concatenated is still readable in all cases. I suppose more than one event per batch doesn't add much coverage, but certainly we need to test multiple batches in order to ensure that the SequenceFile framing is working properly across multiple sync() calls on the same file.

Understood, thanks for clarifying.

kevinconaway:
@mpercy updated the test per your comments.

mpercy:
+1, looks good. I have some style nits (checkstyle is failing) and the merge between master and trunk is a little weird but I will fix those on commit.

PR #59, opened 2016-07-29
taildirsource bug fix: close the file before reopen it if the file is renamed

mlmhl:
In taildirsource module, if a file is renamed, if will be reopened, in this scenario, the old file should be closed before open a new one, otherwise resource leak will occur.


mpercy:
Hi, thanks for the patch. Is it possible to write a test to verify this fix?

mlmhl:
Hi, thanks for good advise, I will add a test soon.

PR #60, opened 2016-08-04
FLUME-2911. Added include pattern option in SpoolDir source

andrearota:
Current implementation of `SpoolDirectorySource` does not allow users to specify a regex pattern to select which files should be monitored. Instead, the current implementation allows users to specify which should not monitored, via the `ignorePattern` parameter.

We implemented the feature, allowing users to specify the include pattern as `a1.sources.src-1.includePattern=^foo.*$` (includes all the files that starts in "foo").

By default, the `includePattern` regex is set to `^.*$` (all files).

Include and exclude patterns can be used at same time and the results are combined.

We also opened this JIRA issue: https://issues.apache.org/jira/browse/FLUME-2911

### DIFF
--- a/flume-ng-core/src/test/java/org/apache/flume/client/avro/TestReliableSpoolingFileEventReader.java
+++ b/flume-ng-core/src/test/java/org/apache/flume/client/avro/TestReliableSpoolingFileEventReader.java
@@ -21,7 +21,9 @@ import com.google.common.base.Charsets;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.io.Files;
+
 import junit.framework.Assert;
+
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.SystemUtils;
 import org.apache.flume.Event;
@@ -113,7 +115,6 @@ public class TestReliableSpoolingFileEventReader {
     }
   }
 
-  
   @Test
   public void testIncludePattern() throws IOException {
     ReliableEventReader reader = new ReliableSpoolingFileEventReader.Builder()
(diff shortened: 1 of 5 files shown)


andrearota:
As requested on [JIRA FLUME-2911](https://issues.apache.org/jira/browse/FLUME-2911), I documented what happens when ignorePattern and includePattern are combined and both match for a given file in the spooling directory.

andrearota:
Added two tests to stress what happens in case both the options are declared and they go in conflict each other, according to [original JIRA request](https://issues.apache.org/jira/browse/FLUME-2911).

flume-ng-core/src/test/java/org/apache/flume/client/avro/TestReliableSpoolingFileEventReader.java
```diff
@@ -112,6 +114,40 @@ private void deleteDir(File dir) {
...
+    ReliableEventReader reader = new ReliableSpoolingFileEventReader.Builder()
+        .spoolDirectory(WORK_DIR)
+        .includePattern("^file2$")
+        .deletePolicy(DeletePolicy.IMMEDIATE.toString())
+        .build();
+ 
+    List<File> before = listFiles(WORK_DIR);
+    Assert.assertEquals("Expected 5, not: " + before, 5, before.size());
+
+    List<Event> events;
```
adenes, line 121: There are a couple of copy-paste lines here and there, could you please refactor it a little bit?
e.g. extract the `reader.readEvents-reader.commit` loops to a method.
The assertion part (`listFiles`, checking the count, removing from the `expectedLeftFiles`, checking the emptiness of it) might be possible to extract too, to improve the readability of the code. (e.g. `void assertExpectedLeftFiles(String... expectedFileNames)`)

flume-ng-core/src/test/java/org/apache/flume/client/avro/TestReliableSpoolingFileEventReader.java
```diff
@@ -137,7 +173,92 @@ public void testIgnorePattern() throws IOException {
...
+    do {
+      events = reader.readEvents(10);
+      reader.commit();
+    } while (!events.isEmpty());
 
+
+    List<File> after = listFiles(WORK_DIR);
+    Assert.assertEquals("Expected 4, not: " + after, 4, after.size());
+        Set<String> expectedLeftFiles = new HashSet<String>(
```
adenes: Wrong indentation, it should be in line with the `Assert` above.
Note: we use 2 spaces for indentation.
You can verify whether your code conforms to Flume's code formatting guidelines by running checkstyle on it: `mvn clean verify -DskipTests -Drat.skip`, please run it, there are a couple of other warnings too.

andrearota: Thank you @adenes for your suggestions. I have refactored the test code and fixed code style violations.

bessbd:
👍

adenes:
lgtm too, 👍

simonati:
looks good to me +1

simonati (approved):

adenes (approved):

mpercy:
+1

mpercy:
This has a merge conflict with trunk but I am going to manually resolve the conflict and squash the PR into a single commit before pushing to trunk.

PR #62, opened 2016-08-16
bugfix: when log rotate, taildir source will treat it as new file, so…

whitelilis:
… duplicate events


asfgit:

iijima-satoshi:
This problem looks like the same as [#291](https://github.com/apache/flume/pull/291).

PR #65, opened 2016-10-11
Kafka Source in combination with Kafka Sink leads to infinite loop

rolandjohann:
Because Kafka Source sets some header fields - especially "topic" - this leads to unexpected behavior in combination with the Kafka Sink => The messages will be consumed and produced again into that topic until the agent will be shut down.

The Kafka Sink determines the topic to which the event should be produced to based on a header field "topic" and overwrites the configuration of the sink.
To (nearly) fix that there are multiple options:
- update the documentation - done with this PR
- explicitly allow topic overriding at the kafka sink via a flag
- write the source topic into another header field - maybe "kafka.topic"


asfgit:

PR #66, opened 2016-10-11
Remove checkstyle from flume-1.7 branch

mpercy:
Something about the checkstyle Maven config is causing the source
artifact build to fail. Let's just remove checkstyle from the 1.7 branch
for the moment until we figure out how to solve this problem. This
should unblock the release.

I tested this and the source artifact appears to be being generated
after applying this patch.


adenes (approved):
👍 thanks for fixing this issue Mike.
I'm happy to work on finding the root cause of it.

bessbd (approved):
LGTM, +1

I'm about to commit this

PR #67, opened 2016-10-11
Add flume-shared module to assembly xmls

adenes:
Fixes the issue that the assembly plugin created the source package without the `flume-shared` project in it so maven compile failed on that.

### DIFF
--- a/flume-ng-dist/src/main/assembly/src.xml
+++ b/flume-ng-dist/src/main/assembly/src.xml
@@ -50,6 +50,7 @@
         <include>org.apache.flume:flume-ng-embedded-agent</include>
         <include>org.apache.flume:flume-tools</include>
         <include>org.apache.flume:flume-ng-auth</include>
+        <include>org.apache.flume:flume-shared</include>
       </includes>
 
       <sources>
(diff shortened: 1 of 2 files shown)


flume-ng-dist/src/main/assembly/src.xml
```diff
@@ -50,6 +50,7 @@
         <include>org.apache.flume:flume-ng-embedded-agent</include>
         <include>org.apache.flume:flume-tools</include>
         <include>org.apache.flume:flume-ng-auth</include>
+        <include>org.apache.flume:flume-shared</include>
```
mpercy, line 53: Should we also add this line to the `<fileSets> <exclude>` section in this file?

adenes: I tried it, there is no difference regardless of whether it's excluded or not. Moreover I discovered that removing all the project dirs from the exclude list has no impact on the files in the src package. I suppose `**/target/**` does the job.

mpercy: Ok thanks Denes! LGTM

mpercy:
+1 from me

bessbd (approved):
LGTM, +1

I'm about to commit this

PR #70, opened 2016-10-12
Remove test dependencies from binary release artifact

mpercy:
This patch removes some test-specific dependencies from the binary
release artifact. These were introduced by the new
flume-shared-kafka-test module that is intended for sharing test code.
Please see the new comment in bin.xml for more information.


bessbd (approved):
LGTM, +1

PR #72, opened 2016-10-14
Fix startup order for ExecSource

mpercy:
This patch starts the SourceCounter in the ExecSource before starting
the runner thread that processes input from the forked command.

Starting the SourceCounter before calling its methods is required to get
correct metric counts because the metrics are all reset to 0 when
SourceCounter.start() is called.

A follow-up patch will attempt to enforce that a SourceCounter or any
other MonitoredCounterGroup is running at the time that calls to modify
the metrics occur.

After applying this patch, TestExecSource.testMonitoredCounterGroup no
longer fails.

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
@@ -165,22 +165,19 @@ public class ExecSource extends AbstractSource implements EventDrivenSource, Con
 
   @Override
   public void start() {
-    logger.info("Exec source starting with command:{}", command);
+    logger.info("Exec source starting with command: {}", command);
+
+    // Start the counter before starting any threads that may access it.
+    sourceCounter.start();
 
     executor = Executors.newSingleThreadExecutor();
+    runner = new ExecRunnable(shell, command, getChannelProcessor(), sourceCounter, restart,
+                              restartThrottle, logStderr, bufferCount, batchTimeout, charset);
 
-    runner = new ExecRunnable(shell, command, getChannelProcessor(), sourceCounter,
-        restart, restartThrottle, logStderr, bufferCount, batchTimeout, charset);
-
-    // FIXME: Use a callback-like executor / future to signal us upon failure.
+    // Start the runner thread.
     runnerFuture = executor.submit(runner);
 
-    /*
-     * NB: This comes at the end rather than the beginning of the method because
-     * it sets our state to running. We want to make sure the executor is alive
-     * and well first.
-     */
-    sourceCounter.start();
+    // Mark the Source as RUNNING.
     super.start();
 
     logger.debug("Exec source started");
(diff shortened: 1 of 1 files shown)


flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
```diff
@@ -165,30 +165,27 @@
...
+    // Start the counter before starting any threads that may access it.
+    sourceCounter.start();
 
-    runner = new ExecRunnable(shell, command, getChannelProcessor(), sourceCounter,
-        restart, restartThrottle, logStderr, bufferCount, batchTimeout, charset);
+    executor = Executors.newSingleThreadExecutor();
+    runner = new ExecRunnable(shell, command, getChannelProcessor(), sourceCounter, restart,
+                              restartThrottle, logStderr, bufferCount, batchTimeout, charset);
 
-    // FIXME: Use a callback-like executor / future to signal us upon failure.
```
simonati, line 175: I guess it is still a TODO informing the rest of flume that ExecSource stopped unexpectedly.

simonati (commented):
It all looks good to me and indeed fix TestExecSource.testMonitoredCounterGroup. (In my env it was constantly failing before)
I only have a nit for the comment. Otherwise all look good to me.

simonati:
(I should have said that it compiles: mvn clean install -DskipTests)

flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
mpercy: I don't really understand the purpose of this, actually I don't think it makes any sense because the startup is async and start() only fails synchronously. There isn't really another mechanism that I know of to "inform the rest of flume that ExecSource stopped unexpectedly". Or is there, and I just don't know about it?

simonati: I'm vaguely guessing that the original comment aimed to take a note about a behavior - which is also specified in the user guide as well - that during startup of ExecSource if the executed command fails it will do that "silently" instead of and this is a TODO.

Example config:

> tier1.sources = source1
> tier1.channels = channel1
> tier1.sinks = sink1
> tier1.channels.channel1.type = file
> tier1.channels.channel1.capacity = 10000
> tier1.sinks.sink1.channel = channel1
> tier1.sources.source1.channels = channel1
> tier1.sources.source1.type = exec
> tier1.sources.source1.command = tail -F /tmp/dirwhichdoesntexist/input.log
> tier1.sinks.sink1.type = logger

Since the user guide also mentions it I'm fine with removing the TODO comment.

mpercy: Okay thanks. Yeah I feel that we should only put actionable TODOs in the code.

simonati (approved):
LGTM

bessbd (approved):
+1, LGTM.

This change indeed fixes the flakiness of the test mentioned in the commit message.

bessbd:
I'm about to commit this

PR #75, opened 2016-10-22
Set MAVEN_OPTS for Travis-CI

bessbd:
Travis builds occasionally fail with Java heap space error.

This commit sets higher limits to fix the issue mentioned above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,9 @@
 # under the License.
 
 language: java
-cache:
-  directories:
-    - $HOME/.m2
 before_install:
-  - echo "MAVEN_OPTS='-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m'" > ~/.mavenrc
+  - export MAVEN_SKIP_RC="true"
 install:
   -
 script:
-  - mvn clean install -DskipTests
+  - MAVEN_OPTS="-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m" mvn clean install -DskipTests
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ cache:
   directories:
     - $HOME/.m2
 before_install:
-  - echo "MAVEN_OPTS='-Xmx2g -XX:MaxPermSize=512m'" > ~/.mavenrc
+  - echo "MAVEN_OPTS='-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m'" > ~/.mavenrc
 install:
   -
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,12 @@
 # under the License.
 
 language: java
-script: mvn clean install -DskipTests
+cache:
+  directories:
+    - $HOME/.m2
+before_install:
+  - echo "MAVEN_OPTS='-Xmx2g -XX:MaxPermSize=512m'" > ~/.mavenrc
+install:
+  -
+script:
+  - mvn clean install -DskipTests
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@
 
 language: java
 before_install:
-  - export MAVEN_SKIP_RC="true"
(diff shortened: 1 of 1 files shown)


liorze:
We actually recommend the following setup: `export MAVEN_OPTS="-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m"`. Have you experienced any issues with this configuration?

Do you think we should enable the tests after we have fixed our flaky tests?

bessbd:
@liorze : thank you for the comment!

I actually copied it from my local environment. That's what I have been using.
(Btw, it looks like to me we aren't exactly consistent with these options. See DEVNOTES vs. README.md)

Regarding the tests: yes, that's what I think.
I see two options:
We either fix all the flakiness and make sure it all fits into reasonable time limits.
Or, we create different test suites, one for smoke testing and another one which contains all the tests we have. (Because some of them aren't exactly "unit" tests, afaik.)

liorze:
I'll update the DEVNOTES as part of FLUME-2945, thanks.

Re/ tests, I think we have a few more tests to fix, it looks doable. Let's give it a try and go with plan b if it takes to much time to fix.

bessbd:
@liorze : thank you for offering to fix the DEVNOTES. (Although, I'm not sure how this is related to switching to Java 8)

Do you have any other review comments for this change?

liorze:
It's related because the `MaxPermSize` option is deprecated in Java 8.

No, LGTM.

bessbd:
I think, if we have MAVEN_OPTS = "..." mvn, then we need to have MAVEN_SKIP_RC = 1.

If we are going to use multiple mvn ... commands, then I think it's easier to set it up in one place. (before_install)

The caching is to speed up the builds. ( https://docs.travis-ci.com/user/caching/#Arbitrary-directories )
I can remove it, that's fine.

mpercy:
> The caching is to speed up the builds.

Thanks for helping me understand. Maybe add a comment about this for people new to Travis?

> I think, if we have MAVEN_OPTS = "..." mvn, then we need to have MAVEN_SKIP_RC = 1.

Why? I don't understand the reason

> If we are going to use multiple mvn ... commands, then I think it's easier to set it up in one place. 

We aren't doing that yet, right? Would we ever do that?

Personally I like to avoid putting stuff in default locations for build scripts in order to avoid unintended side effects either in other concurrent builds or in future builds. I just don't like to "pollute" the environment. However, I'm guessing Travis uses a Docker image or something like that and wipes out the home dir before each build so you can leave cruft in $HOME and it will never affect future builds. Is that right?

bessbd:
> Thanks for helping me understand. Maybe add a comment about this for people new to Travis?

I did not notice **huge** speedups, so I'm just going to remove it for now.

> Why? I don't understand the reason

According to http://stackoverflow.com/a/35755762/5323166 , if there are any defaults (mavenrc, etc.) set in the FS, `MAVEN_OPTS` env var is ignored.

Afaik, all Travis-CI build environments are "disposable" (the Docker stuff you have mentioned), so it should be fine to "pollute" the environment, but we can go YAGNI and change it later.

.travis.yml
```diff
@@ -16,12 +16,9 @@
 # under the License.
 
 language: java
-cache:
-  directories:
-    - $HOME/.m2
 before_install:
-  - echo "MAVEN_OPTS='-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m'" > ~/.mavenrc
+  - export MAVEN_SKIP_RC="true"
```
mpercy: Do you have evidence that this line is required? If so, can you please add a comment about it?

mpercy:
> I did not notice huge speedups, so I'm just going to remove it for now.

I wasn't asking to remove it, just for a comment to be added to the yaml file, but either way is OK with me.

.travis.yml
bessbd: When I remove the MAVEN_SKIP_RC setting, the build fails.

I think, that line is required. I'll include a comment about this.

mpercy (approved):
+1 looks good, thanks!

bessbd:
Thank you, @mpercy 

I'm about to commit this

PR #76, opened 2016-10-23
a EventDeserializer for spooldir

cjn082030:
A simple EventDeserializer for spooldir.A deserializer that parses text lines from a file.Matching line/multiLine against the regular expression.


asfgit:

PR #82, opened 2016-10-25
[WIP] Proof of concept for a generic password provider

adenes:
One of the main security concerns regarding Flume is that currently passwords can only be set in plain text in the config file. I have a proof-of-concept to overcome this limitation with an extensible password provider.

The core of the solution is the `PasswordProvider` interface which has a default implementation (`PlainTextPasswordProvider`) which returns the value of the given key, thus taking care of backwards compatibility.
The other implementation is the `ExternalProcessPasswordProvider` which executes the configured command and returns its output.

Usage example can be seen in the `AvroSource` (see the 2nd commit of this PR):

``` diff
- keystorePassword = context.getString(KEYSTORE_PASSWORD_KEY);
+ keystorePassword = PasswordConfigurator.getPassword(context, KEYSTORE_PASSWORD_KEY);
```

**Example configuration to use the `ExternalProcessPasswordProvider`:**

```
...
a.sources.avro.keystore-password.passwordProviderClass=org.apache.flume.conf.ExternalProcessPasswordProvider
a.sources.avro.keystore-password.command=get_avro_keystore_password.sh
...
```

**Example configuration with no `passwordProviderClass` set:**

```
...
a.sources.avro.keystore-password=SecretPassword
...
```

As no `passwordProviderClass` is set in this example the default `PlainTextPasswordProvider` is used which returns the value of `a.sources.avro.keystore-password`.

---

_Note: this is still a work in progress, I wanted to sketch up my idea. Any questions/comments/suggestions are more than welcome._

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/conf/ExternalProcessPasswordProvider.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/conf/ExternalProcessPasswordProvider.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.flume.conf;
+
+import com.google.common.base.Optional;
+import com.google.common.io.ByteStreams;
+import org.apache.flume.Context;
+
+import java.io.IOException;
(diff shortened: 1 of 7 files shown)


flume-ng-core/src/main/java/org/apache/flume/conf/ExternalProcessPasswordProvider.java
```diff
@@ -0,0 +1,89 @@
...
+  public String getPassword(Context context, String key) {
+    if (!password.isPresent()) {
+      password = Optional.of(loadPassword(context, key));
+    }
+    return password.get();
+  }
+
+  private String loadPassword(Context context, String key) {
+    String charsetKey = key + "." + CHARSET_KEY;
+    String charsetName = context.getString(charsetKey, "iso-8859-1");
```
bessbd, line 46: Why iso-8859-1?

adenes: As it's about the output of a shell script I don't expect to have non-ascii characters, that's why I decided to use iso-8859-1 by default. I might be wrong so I'm open to reconsider this.

mpercy:
I didn't look at this much yet, but would this ultimately be compatible with standard JAAS credential providers, like the one Hadoop implements?

flume-ng-core/src/main/java/org/apache/flume/conf/ExternalProcessPasswordProvider.java
simonati: +1  for Unicode standard and UTF-8 implementation even for defaults.

asfgit:

PR #83, opened 2016-10-28
FLUME-2812: Exception in thread "SinkRunner-PollingRunner-DefaultSinkProcessor" java.lang.Error: Maximum permit count exceeded

adenes:
Fix semaphore leak causing `java.lang.Error: Maximum permit count exceeded` in `MemoryChannel`

``` java
Exception in thread "SinkRunner-PollingRunner-DefaultSinkProcessor" java.lang.Error: Maximum permit count exceeded
    at java.util.concurrent.Semaphore$Sync.tryReleaseShared(Semaphore.java:192)
    at java.util.concurrent.locks.AbstractQueuedSynchronizer.releaseShared(AbstractQueuedSynchronizer.java:1341)
    at java.util.concurrent.Semaphore.release(Semaphore.java:609)
    at org.apache.flume.channel.MemoryChannel$MemoryTransaction.doCommit(MemoryChannel.java:147)
    at org.apache.flume.channel.BasicTransactionSemantics.commit(BasicTransactionSemantics.java:151)
    at org.apache.flume.sink.AbstractRpcSink.process(AbstractRpcSink.java:379)
    at org.apache.flume.sink.DefaultSinkProcessor.process(DefaultSinkProcessor.java:68)
    at org.apache.flume.SinkRunner$PollingRunner.run(SinkRunner.java:147)
    at java.lang.Thread.run(Thread.java:745)
```

for more details and explanation see my comment on [FLUME-2812](https://issues.apache.org/jira/browse/FLUME-2812)

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/channel/MemoryChannel.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/channel/MemoryChannel.java
@@ -18,6 +18,7 @@
  */
 package org.apache.flume.channel;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.flume.ChannelException;
 import org.apache.flume.ChannelFullException;
@@ -171,7 +172,6 @@ public class MemoryChannel extends BasicChannelSemantics {
         }
         putList.clear();
       }
-      bytesRemaining.release(putByteCounter);
       putByteCounter = 0;
       takeByteCounter = 0;
 
@@ -373,5 +373,10 @@ public class MemoryChannel extends BasicChannelSemantics {
     }
     //Each event occupies at least 1 slot, so return 1.
     return 1;
+  }
+
+  @VisibleForTesting
+  int getBytesRemainingValue() {
+    return bytesRemaining.availablePermits();
   }
 }
--- a/flume-ng-core/src/main/java/org/apache/flume/channel/MemoryChannel.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/channel/MemoryChannel.java
@@ -54,7 +54,10 @@ public class MemoryChannel extends BasicChannelSemantics {
   private static Logger LOGGER = LoggerFactory.getLogger(MemoryChannel.class);
   private static final Integer defaultCapacity = 100;
   private static final Integer defaultTransCapacity = 100;
-  private static final double byteCapacitySlotSize = 100;
+
+  @VisibleForTesting
+  static final double byteCapacitySlotSize = 100;
+
   private static final Long defaultByteCapacity = (long)(Runtime.getRuntime().maxMemory() * .80);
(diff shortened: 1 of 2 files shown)


bessbd:

Is this PR related to https://issues.apache.org/jira/browse/FLUME-2461 ?

(Or maybe https://issues.apache.org/jira/browse/FLUME-2739 ?)

adenes:
Both tickets are related but none of those fixes the issue releasing non-acquired permits of the `bytesRemaining` semaphore.

bessbd:
@adenes : I see now.

Do you have any easy way to reproduce this error?

adenes:
@bessbd: I've created a test case explicitly for the _Maximum permit count exceeded_ error: `testAvoidMaximumPermitCountExceeded()`. It fails with the not-fixed version of the `MemoryChannel`.
I'm not sure whether it should be merged too, because this error is just a symptom of the leaking semaphores, which is covered in the `testByteCapacityAfterRollback`.

flume-ng-core/src/main/java/org/apache/flume/channel/MemoryChannel.java
```diff
@@ -171,7 +175,6 @@ protected void doRollback() {
         }
         putList.clear();
       }
-      bytesRemaining.release(putByteCounter);
```
simonati, line 174: +1 on this change. It solves the problem of leaking semaphore caused by rollBacks. (Essentially rollback should be a "purely drop all temporal changes" and commit should be a "check if temporal changes are applicable then apply")

Additionally please note that doCommit is still able to leak as acquiring the queueRemaining and bytesRemaining semaphores are outside of the synchronised block. (Purely moving them inside the block would have a huge perf penalty in situations when channel is nearly full and keepAlive was set) I would recommend a two level check (one without wait within the synchronised block and another upfront check which could wait)  in a separate change.

simonati (approved):
LGTM

mvn test -pl flume-ng-core only reported unrelevant (Avro,Thrift) failed tests

bessbd (approved):
+1, LGTM

I'm about to commit this

bessbd:
Please, note that I'll only merge the first commit. See https://github.com/apache/flume/pull/83#issuecomment-258153972 for further information

bessbd:
@satiatcloudera : thank you for the review!

skt0001:
Hi, I am getting below error when reading logs from rabbitmq server to HDFS path using flume
 
 
 
Exception in thread "PollableSourceRunner-RabbitMQSource-rabbitmq-source1" java.lang.AbstractMethodError: org.apache.flume.source.rabbitmq.RabbitMQSource.getBackOffSleepIncrement()J
at org.apache.flume.source.PollableSourceRunner$PollingRunner.run(PollableSourceRunner.java:142)
at java.lang.Thread.run(Thread.java:748)

PR #85, opened 2016-11-16
FLUME-3025: Expose FileChannel.open JMX

adenes:
This patch exposes the `FileChannel`'s `open` flag on JMX to make it possible to detect when it wasn't able to start up (thus its `open` flag is `false`).

### DIFF
--- a/flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
+++ b/flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
@@ -40,6 +40,7 @@ import org.apache.flume.channel.file.encryption.EncryptionConfiguration;
 import org.apache.flume.channel.file.encryption.KeyProvider;
 import org.apache.flume.channel.file.encryption.KeyProviderFactory;
 import org.apache.flume.instrumentation.ChannelCounter;
+import org.apache.flume.channel.file.instrumentation.FileChannelCounter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -267,13 +268,14 @@ public class FileChannel extends BasicChannelSemantics {
     }
 
     if (channelCounter == null) {
-      channelCounter = new ChannelCounter(getName());
+      channelCounter = new FileChannelCounter(this);
     }
   }
 
   @Override
   public synchronized void start() {
     LOG.info("Starting {}...", this);
+    channelCounter.start();
     try {
       Builder builder = new Log.Builder();
       builder.setCheckpointInterval(checkpointInterval);
@@ -312,7 +314,6 @@ public class FileChannel extends BasicChannelSemantics {
       }
     }
     if (open) {
-      channelCounter.start();
       channelCounter.setChannelSize(getDepth());
       channelCounter.setChannelCapacity(capacity);
     }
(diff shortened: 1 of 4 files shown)


flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/instrumentation/FileChannelCounter.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.flume.channel.file.instrumentation;
+
+import org.apache.flume.channel.file.FileChannel;
+import org.apache.flume.instrumentation.ChannelCounter;
+
+/**
```
simonati: Shouldn't license be the first thing of every source file? If yes than it is interesting that it passed RAT check.

adenes: You are right, thanks, will commit the fix soon.

flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/instrumentation/FileChannelCounter.java
```diff
@@ -0,0 +1,37 @@
...
+public class FileChannelCounter extends ChannelCounter implements FileChannelCounterMBean {
+
+  private final FileChannel fileChannel;
+
+  public FileChannelCounter(FileChannel fileChannel) {
+    super(fileChannel.getName());
+    this.fileChannel = fileChannel;
+  }
+
+  public boolean isOpen() {
```
simonati, line 32: I guess the idea behind the Counters is that they are not on the critical path, the object which is monitored can offer the information to them instead of pulled by the Counter. I guess this would be dangerous the other way around: eg if the counter fetches the size of the channel that might block the actual workflow (so a monitoring UI would effectively block dataflow). In this change I see the same pattern the counter pulls the isOpen information from the filechannel. In this particular case the isOpen doesn't block but it in the future this assumption might fail.

adenes: Thank you for your review. I wasn't sure which way to go, because I didn't want to put the responsibility of updating the flag to the FileChannel. (Which shouldn't be a problem, to be honest, as all the other counters work like this).
Thanks for the good example, I agree with you and I'll update my patch according to your comment.

flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
```diff
@@ -401,6 +402,18 @@ public boolean isOpen() {
   }
 
   /**
+   * This method makes sure that <code>this.open</code> and <code>channelCounter.open</code>
```
simonati, line 405: Thanks for the javadoc! It is very useful.

simonati (approved):
LGTM

adenes:
Thank you for the reviews @satiatcloudera.
@bessbd, @jholoman, @mpercy could you please do a final review? Thanks.

bessbd (approved):
LGTM

bessbd:
I'm about to commit this

bessbd:

bessbd:
@satiatcloudera : thank you for the review!

PR #87, opened 2016-11-18
Remove `hostHeader = hostname` property from Host interceptor example

peterableda:
We are overriding the host header name from `host` to `hostname` in the example usage section. Due to this example users are overriding the header name too but still use the `%{host}` substitution as shown in the HDFS Sink section. This won't work for them.

This change removes this config line.


bessbd:
+1, LGTM

I'll leave some time for others to review this, then commit it if nobody disagrees.

bessbd:
I'm about to commit this.

bessbd:
@peterableda : thank you for the patch!

PR #92, opened 2016-11-23
FLUME-3027 Kafka Channel should clear offsets map after commit

jholoman:
Added a call to clear the offsets map after a commit so as to avoid repeatedly committing already-committed offsets. Also updated various debug and trace log messages/calls to help with troubleshooting.

### DIFF
--- a/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
+++ b/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
@@ -484,7 +484,7 @@ public class KafkaChannel extends BasicChannelSemantics {
         e = consumerAndRecords.get().failedEvents.removeFirst();
       } else {
         logger.trace("Assignment during take: {}",
-           consumerAndRecords.get().consumer.assignment().toString());
+            consumerAndRecords.get().consumer.assignment().toString());
         try {
           long startTime = System.nanoTime();
           if (!consumerAndRecords.get().recordIterator.hasNext()) {
@@ -568,8 +568,7 @@ public class KafkaChannel extends BasicChannelSemantics {
           consumerAndRecords.get().commitOffsets();
           long endTime = System.nanoTime();
           counter.addToKafkaCommitTimer((endTime - startTime) / (1000 * 1000));
-          logger.debug("Latest committed offsets: {}",
-              consumerAndRecords.get().getCommittedOffsetsString());
+          logger.debug("{}", consumerAndRecords.get().getCommittedOffsetsString());
         }
 
         int takes = events.get().size();
@@ -723,7 +722,7 @@ public class KafkaChannel extends BasicChannelSemantics {
               .append(this.consumer.committed(tp).offset())
               .append("] ");
         } catch (NullPointerException npe) {
-            logger.debug("Committed {}", tp);
+          logger.debug("Committed {}", tp);
         }
       }
       return sb.toString();
--- a/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
+++ b/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
@@ -154,8 +154,7 @@ public class KafkaChannel extends BasicChannelSemantics {
     producer.close();
     counter.stop();
     super.stop();
-    logger.info("Kafka channel {} stopped. Metrics: {}", getName(),
-            counter);
+    logger.info("Kafka channel {} stopped.", getName());
   }
 
   @Override
@@ -372,6 +371,7 @@ public class KafkaChannel extends BasicChannelSemantics {
   }
 
   private void decommissionConsumerAndRecords(ConsumerAndRecords c) {
+    c.consumer.wakeup();
     c.consumer.close();
   }
 
@@ -564,23 +564,25 @@ public class KafkaChannel extends BasicChannelSemantics {
                   ex);
         }
       } else {
-        //event taken ensures that we have collected events in this transaction
+        // event taken ensures that we have collected events in this transaction
         // before committing
         if (consumerAndRecords.get().failedEvents.isEmpty() && eventTaken) {
-          logger.trace("about to commit batch.");
+          logger.trace("About to commit batch.");
           long startTime = System.nanoTime();
           consumerAndRecords.get().commitOffsets();
           long endTime = System.nanoTime();
           counter.addToKafkaCommitTimer((endTime - startTime) / (1000 * 1000));
-          counter.addToEventTakeSuccessCount(Long.valueOf(events.get().size()));
 
           if (logger.isDebugEnabled()) {
             logger.debug(consumerAndRecords.get().getCommittedOffsetsString());
           }
         }
-        logger.trace("clearing offsets map.");
-        consumerAndRecords.get().offsets.clear();
-        events.get().clear();
+
+        int takes = events.get().size();
+        if (takes > 0) {
+          counter.addToEventTakeSuccessCount(takes);
+          events.get().clear();
+        }
       }
     }
 
@@ -593,7 +595,7 @@ public class KafkaChannel extends BasicChannelSemantics {
         producerRecords.get().clear();
         kafkaFutures.get().clear();
       } else {
-        counter.addToRollbackCounter(Long.valueOf(events.get().size()));
+        counter.addToRollbackCounter(events.get().size());
         consumerAndRecords.get().failedEvents.addAll(events.get());
         events.get().clear();
       }
@@ -685,17 +687,19 @@ public class KafkaChannel extends BasicChannelSemantics {
     }
 
     void poll() {
-      logger.trace("polling with timeout: {}ms channel-{}", pollTimeout, channelUUID);
+      logger.trace("polling with timeout: {}ms channel-{}", pollTimeout, getName());
       this.records = consumer.poll(pollTimeout);
       this.recordIterator = records.iterator();
-      logger.debug("returned {} records from last poll channel-{}", records.count(), channelUUID);
+      logger.debug("returned {} records from last poll channel-{}", records.count(), getName());
     }
 
     void commitOffsets() {
       this.consumer.commitSync(offsets);
+      logger.trace("About to clear offsets map.");
+      this.offsets.clear();
     }
 
-    String getOffsetMapString() {
+    private String getOffsetMapString() {
       StringBuilder sb = new StringBuilder();
       sb.append("Current offsets map: ");
       for (TopicPartition tp : offsets.keySet()) {
@@ -706,7 +710,7 @@ public class KafkaChannel extends BasicChannelSemantics {
     }
 
     // This prints the current committed offsets when debug is enabled
-    String getCommittedOffsetsString() {
+    private String getCommittedOffsetsString() {
       StringBuilder sb = new StringBuilder();
       sb.append("Channel ").append(this.uuid)
           .append(" committed: ");
--- a/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
+++ b/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
@@ -103,9 +103,9 @@ public class KafkaChannel extends BasicChannelSemantics {
   private Integer staticPartitionId;
   private boolean migrateZookeeperOffsets = DEFAULT_MIGRATE_ZOOKEEPER_OFFSETS;
 
-  //used to indicate if a rebalance has occurred during the current transaction
+  // used to indicate if a rebalance has occurred during the current transaction
   AtomicBoolean rebalanceFlag = new AtomicBoolean();
-  //This isn't a Kafka property per se, but we allow it to be configurable
+  // This isn't a Kafka property per se, but we allow it to be configurable
   private long pollTimeout = DEFAULT_POLL_TIMEOUT;
 
 
@@ -166,7 +166,7 @@ public class KafkaChannel extends BasicChannelSemantics {
   @Override
   public void configure(Context ctx) {
 
-    //Can remove in the next release
+    // Can remove in the next release
     translateOldProps(ctx);
 
     topicStr = ctx.getString(TOPIC_CONFIG);
@@ -217,7 +217,7 @@ public class KafkaChannel extends BasicChannelSemantics {
       logger.warn("{} is deprecated. Please use the parameter {}", "topic", TOPIC_CONFIG);
     }
 
-    //Broker List
+    // Broker List
     // If there is no value we need to check and set the old param and log a warning message
     if (!(ctx.containsKey(BOOTSTRAP_SERVERS_CONFIG))) {
       String brokerList = ctx.getString(BROKER_LIST_FLUME_KEY);
@@ -230,7 +230,7 @@ public class KafkaChannel extends BasicChannelSemantics {
       }
     }
 
-    //GroupId
+    // GroupId
     // If there is an old Group Id set, then use that if no groupId is set.
     if (!(ctx.containsKey(KAFKA_CONSUMER_PREFIX + ConsumerConfig.GROUP_ID_CONFIG))) {
       String oldGroupId = ctx.getString(GROUP_ID_FLUME);
@@ -265,7 +265,7 @@ public class KafkaChannel extends BasicChannelSemantics {
     producerProps.put(ProducerConfig.ACKS_CONFIG, DEFAULT_ACKS);
     producerProps.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, DEFAULT_KEY_SERIALIZER);
     producerProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, DEFAULT_VALUE_SERIAIZER);
-    //Defaults overridden based on config
+    // Defaults overridden based on config
     producerProps.putAll(ctx.getSubProperties(KAFKA_PRODUCER_PREFIX));
     producerProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, bootStrapServers);
   }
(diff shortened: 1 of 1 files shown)


flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -564,14 +564,22 @@ protected void doCommit() throws InterruptedException {
                   ex);
         }
       } else {
+        //event taken ensures that we have collected events in this transaction
```
simonati: nit: missing space after //

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -564,14 +564,22 @@ protected void doCommit() throws InterruptedException {
...
+        // before committing
         if (consumerAndRecords.get().failedEvents.isEmpty() && eventTaken) {
```
simonati, line 566: Naive question: could you please tell why the else path is not needed. Condition to the if statement suggests that we silently ignore some failure scenario.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -677,34 +685,45 @@ private Event deserializeValue(byte[] value, boolean parseAsFlumeEvent) throws I
...
+    String getCommittedOffsetsString() {
       StringBuilder sb = new StringBuilder();
+      sb.append("Channel ").append(this.uuid)
+          .append(" committed: ");
       for (TopicPartition tp : this.consumer.assignment()) {
         try {
-          sb.append("Committed: [").append(tp).append(",")
+          sb.append("[").append(tp).append(",")
               .append(this.consumer.committed(tp).offset())
-              .append(",").append(this.consumer.committed(tp).metadata()).append("]");
```
simonati, line 697: is consumer.committed(tp).metadata() not useful anymore?

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -677,34 +685,45 @@ private Event deserializeValue(byte[] value, boolean parseAsFlumeEvent) throws I
...
+      for (TopicPartition tp : offsets.keySet()) {
+        sb.append("p").append(tp.partition()).append("-")
+            .append(offsets.get(tp).offset()).append(" ");
+      }
+      return sb.toString();
+    }
 
-    public void printCurrentAssignment() {
+    // This prints the current committed offsets when debug is enabled
+    String getCommittedOffsetsString() {
```
simonati: I think this should be private

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -677,34 +685,45 @@ private Event deserializeValue(byte[] value, boolean parseAsFlumeEvent) throws I
...
-      logger.trace("polling");
+      logger.debug("returned {} records from last poll channel-{}", records.count(), channelUUID);
     }
 
     void commitOffsets() {
       this.consumer.commitSync(offsets);
     }
 
-    // This will reset the latest assigned partitions to the last committed offsets;
+    String getOffsetMapString() {
```
simonati: I think this should be private

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -564,14 +564,22 @@ protected void doCommit() throws InterruptedException {
...
+        //event taken ensures that we have collected events in this transaction
+        // before committing
         if (consumerAndRecords.get().failedEvents.isEmpty() && eventTaken) {
+          logger.trace("about to commit batch.");
           long startTime = System.nanoTime();
           consumerAndRecords.get().commitOffsets();
           long endTime = System.nanoTime();
           counter.addToKafkaCommitTimer((endTime - startTime) / (1000 * 1000));
-          consumerAndRecords.get().printCurrentAssignment();
+          counter.addToEventTakeSuccessCount(Long.valueOf(events.get().size()));
```
simonati: nit: could you please remove Long.valueOf? it is not needed (here and in the doRollback function)

simonati (changes requested):
1) Log statement changes make sense to me. 

2) I would like to suggest calling the clear from the commitOffsets function within the ConsumerAndRecords class. I think this has to be addressed.

3) I would like to understand why we don't have an ``else`` path in doCommit to cover the ``failure`` scenario.

4) ConsumerAndRecords private inner class has already grown to the size which might worth the consideration of transferring it to a separate class. (As it has been already touched by this change this might be a good time for some cleanup. It won't hurt my feelings if not this time.)

5) The ``clear`` functionality might need a test. I don't know whether kafka provide an easy way to test this seemingly internal behaviour. If I got it correctly it is about giving a halfway filled topicpartition<->offset mappings as argument when calling clear after commit. This essentially translates the state-of-the-world-offset map to a not-yet-committed-offsets map. What do you think?

Haven't run the change yet.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: I haven't found a good use for it, and it's just polluting the logs.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: Agree

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: Sure

jholoman:
Thanks @satiatcloudera. 

2) Totally agree. I had a version with this and somehow it got reverted.
3) process() from a sink calls channel.take() until the batch size. If there is no event we break out of the loop and then commit the transaction. See https://github.com/apache/flume/blob/trunk/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java#L362. 

The logic here is a little complicated, and it's always worth another look. We could probably catch the Exceptions that the Kafka Consumer could throw during commitSync CommitFailedException and Wakeup exception. Is it ok if we address this logic in another jira? 

4) I agree, but I'd like to separate that change. 

5) We are just clearing our internal tracking Map. This essentially prevents committing previously committed offsets. In a topic with a large number of partitions, this can be problematic. 

All of this is kind of hard to follow. I think the unit-tests are kind of tricky to understand, and the logic could be clearer. 

The intent of this patch is to address a situation where the Channel can get totally stuck. Happy to take another pass at refactoring the channel.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: This needs to be moved back out of the if statement.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: See comment

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -497,21 +495,18 @@ protected Event doTake() throws InterruptedException {
             e = deserializeValue(record.value(), parseAsFlumeEvent);
             TopicPartition tp = new TopicPartition(record.topic(), record.partition());
             OffsetAndMetadata oam = new OffsetAndMetadata(record.offset() + 1, batchUUID);
+            //we should refactor this into a method in the subclass
             consumerAndRecords.get().offsets.put(tp, oam);
 
-            if (logger.isTraceEnabled()) {
```
simonati, line 502: Have you measuered the performance impact of this change? Based on general guidance this should be the other way around except perf measurement shows otherwise. I would like to ask you moving back the log level checking condition if you do relatively heavy computation for calculating the argument list for logging function calls on critical paths (doTake is considered as critical path)

Please see: http://logging.apache.org/log4j/1.2/manual.html#performance

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -154,8 +155,7 @@ public void stop() {
     producer.close();
     counter.stop();
     super.stop();
-    logger.info("Kafka channel {} stopped. Metrics: {}", getName(),
```
simonati, line 157: I would like to ask you to revert this line. I consider shutdown metrics to be quite useful.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -497,21 +495,18 @@ protected Event doTake() throws InterruptedException {
             e = deserializeValue(record.value(), parseAsFlumeEvent);
             TopicPartition tp = new TopicPartition(record.topic(), record.partition());
             OffsetAndMetadata oam = new OffsetAndMetadata(record.offset() + 1, batchUUID);
+            //we should refactor this into a method in the subclass
```
simonati: nit: missing whitespace after //

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -482,11 +483,8 @@ protected Event doTake() throws InterruptedException {
       if (!consumerAndRecords.get().failedEvents.isEmpty()) {
         e = consumerAndRecords.get().failedEvents.removeFirst();
       } else {
-
-        if (logger.isDebugEnabled()) {
```
simonati, line 486: Please see my comment on logging performance.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -497,21 +495,18 @@ protected Event doTake() throws InterruptedException {
...
             if (record.key() != null) {
               e.getHeaders().put(KEY_HEADER, record.key());
             }
 
-            if (logger.isDebugEnabled()) {
-              logger.debug("Processed output from partition {} offset {}",
-                           record.partition(), record.offset());
-            }
+            logger.debug("Processed output from partition {} offset {}",
+                record.partition(), record.offset());
```
simonati: This looks good as partition() and offset() only returns a value and performs no computation.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -564,15 +560,22 @@ protected void doCommit() throws InterruptedException {
...
+        // event taken ensures that we have collected events in this transaction
+        // before committing
         if (consumerAndRecords.get().failedEvents.isEmpty() && eventTaken) {
+          logger.trace("About to commit batch.");
           long startTime = System.nanoTime();
           consumerAndRecords.get().commitOffsets();
           long endTime = System.nanoTime();
           counter.addToKafkaCommitTimer((endTime - startTime) / (1000 * 1000));
-          consumerAndRecords.get().printCurrentAssignment();
+          logger.debug("{}", consumerAndRecords.get().getCommittedOffsetsString());
```
simonati: Please see my comment on logging performance.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -373,6 +372,7 @@ private void migrateOffsets() {
   }
 
   private void decommissionConsumerAndRecords(ConsumerAndRecords c) {
+    c.consumer.wakeup();
```
simonati, line 375: What is the intention here?
From the javadoc of KafkaConsumer.close() it seems that wakeup won't interrupt the stop. So I would assume something else should have been interrupted by this statement. In that case wouldn't be the desired order: wakeup first then close?

> org.apache.kafka.clients.consumer.KafkaConsumer
> public void close()
> Close the consumer, waiting indefinitely for any needed cleanup. If auto-commit is enabled, this will commit the current offsets. Note that wakeup() cannot be use to interrupt close.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
simonati: Thanks for the explanation. Fine by me.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -564,15 +560,22 @@ protected void doCommit() throws InterruptedException {
...
+          logger.trace("About to commit batch.");
           long startTime = System.nanoTime();
           consumerAndRecords.get().commitOffsets();
           long endTime = System.nanoTime();
           counter.addToKafkaCommitTimer((endTime - startTime) / (1000 * 1000));
-          consumerAndRecords.get().printCurrentAssignment();
+          logger.debug("{}", consumerAndRecords.get().getCommittedOffsetsString());
+        }
+
+        int takes = events.get().size();
```
simonati, line 577: Have you considered moving this into the 
> if (consumerAndRecords.get().failedEvents.isEmpty() && eventTaken) {

block?

simonati (changes requested):
Thank you for the update! The already discussed items look good these are only findings related to additional changes added after prev review.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: My understanding is that using the {} means they aren't evaluated if the logging mode doesn't match. Am I wrong about this?

http://www.slf4j.org/faq.html#logging_performance

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: They are already printed in the Flume Agent log when the channel closes:

org.apache.flume.instrumentation.MonitoredCounterGroup: Shutdown Metric for type: CHANNEL, name: kafka-channel-1. channel.start.time == 1480359313437

So this was redundant.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: The intention is to break out of the blocking thread in poll(), before closing the consumer.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
jholoman: Maybe? I don't have a strong opinion on it and would prefer to make as few changes as possible.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
simonati: Please ignore my comment I misread this change. It is completely fine in the way you done it originally.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
simonati: You can save the cost of string concatenation with {} if log level doesn't match (which is already a gain). In java before calling any function, its argument list is evaluated. So JVM will execute consumerAndRecords.get().getOffsetMapString() statement before invoking logger.trace(arg1, arg2) to make sure arg2 is computed so that it can be passed into.

simonati: You can save the cost of string concatenation with {} if log level doesn't match (which is already a gain). In java before calling any function, its argument list is evaluated. So JVM will execute consumerAndRecords.get().getOffsetMapString() statement before invoking logger.trace(arg1, arg2) to make sure arg2 is computed so that it can be passed into.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
simonati: Please ignore my prev comment I misread this change. Sorry for that. It is completely fine in the way you done it originally.

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
simonati: Indeed. conter.stop() logs this data at info level already.

bessbd (approved):
LGTM, +1

bessbd:
I'm about to commit this

bessbd:
@jholoman : thank you for the patch!

bessbd:
@simonati : thank you for the reviews!

JerryShaoPro:
I also found this bug when an agent down and after kafka rebalance. After read the code of the commit offsets part, I found offsets map has not been clear, when the cousumer's partitions changed. I'd like to commit a patch to fix this bug, but found it has been fixed by others, haha.

Thanks all of the above persons.

qqj1979:
- KafkaSource EXCEPTION, {}
org.apache.kafka.clients.consumer.CommitFailedException: Commit cannot be completed since the group has already rebalanced and assigned the partitions to another member. This means that the time between subsequent calls to poll() was longer than the configured session.timeout.ms, which typically implies that the poll loop is spending too much time message processing. You can address this either by increasing the session timeout or by reducing the maximum size of batches returned in poll() with max.poll.records.

**I encountered this problem, how to solve the solution。**

PR #93, opened 2016-11-25
DEFAULT_AUTO_OFFSET_RESET should be "latest" according to official do…

hangc0276:
In the official document, the default value of "kafka.consumer.auto.offset.reset" is "latest", but in the source code, "DEFAULT_AUTO_OFFSET_RESET" is "earliest".

### DIFF
--- a/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannelConfiguration.java
+++ b/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannelConfiguration.java
@@ -44,7 +44,7 @@ public class KafkaChannelConfiguration {
 
   public static final String KEY_HEADER = "key";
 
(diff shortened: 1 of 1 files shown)


simonati:
I think the reset configuration for kafkachannel is kind of internal. Without it the channel would be able to guarantee message delivery. With "latest" it would skip a whole bunch of events when offset has to be reseted.

simonati:
I would recommend changing the docs instead of the source

hangc0276:
With "earliest" it would reset the offset when the consumer restarted or rebalanced, which will lead to message repeat consumption

simonati:
Existing implementation achieves at least once message delivery. This doesn't guarantee that messages are delivered more than once (in fact it is a known downside). Exposing this configuration allows users to change the delivery characteristics of kafkachannel - if really needed - but making it default would break existing setups.

jholoman:
Agree that on this we need to document the existing default value.

asfgit:

PR #97, opened 2016-12-03
FLUME-2993. Add support for environment variables in configuration files

bessbd:
Flume does not currently support environment variable interpolation in the properties file configuration.

Enabling it helps with:
* removing security credentials from config files
* copy-pastes in configuration files when defining multiple agents

### DIFF
--- a/flume-ng-doc/sphinx/FlumeUserGuide.rst
+++ b/flume-ng-doc/sphinx/FlumeUserGuide.rst
@@ -234,6 +234,23 @@ The original Flume terminal will output the event in a log message.
 
 Congratulations - you've successfully configured and deployed a Flume agent! Subsequent sections cover agent configuration in much more detail.
 
+Using environment variables in configuration files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If we enable environment variable interpolation by adding `-DpropertiesImplementation=org.apache.flume.node.EnvVarResolverProperties`
+to our `bin/flume-ng` call, we can use environment variables in the config, such as::
+
+  a1.sources = r1
+  a1.sources.r1.type = netcat
+  a1.sources.r1.bind = 0.0.0.0
+  a1.sources.r1.port = ${NC_PORT}
+  a1.sources.r1.channels = c1
+
+So a complete command line start could look like::
+
+  $ NC_PORT=44444 bin/flume-ng agent --conf conf --conf-file example.conf --name a1 -Dflume.root.logger=INFO,console -DpropertiesImplementation=org.apache.flume.node.EnvVarResolverProperties
+
+
 Logging raw data
 ~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -234,6 +234,23 @@ The original Flume terminal will output the event in a log message.
...
+  a1.sources.r1.type = netcat
+  a1.sources.r1.bind = 0.0.0.0
+  a1.sources.r1.port = ${NC_PORT}
+  a1.sources.r1.channels = c1
+
+So a complete command line start could look like::
+
+  $ NC_PORT=44444 bin/flume-ng agent --conf conf --conf-file example.conf --name a1 -Dflume.root.logger=INFO,console -DpropertiesImplementation=org.apache.flume.node.EnvVarResolverProperties
+
+
```
jholoman, line 255: I think we should be explicit about a couple of different things in the user guide

1) This only affects values and not keys
2) You can use env variables however you want, the example is just an example. 

I would also review the language a little bit (eg, remove "we")

flume-ng-node/src/test/java/org/apache/flume/node/TestEnvVarResolverProperties.java
```diff
@@ -0,0 +1,60 @@
...
+    provider = new PropertiesFileConfigurationProvider("a1", TESTFILE);
+  }
+
+  @Test
+  public void resolveEnvVars() throws Exception {
+    environmentVariables.set("VARNAME", "varvalue");
+    String resolved = EnvVarResolverProperties.resolveEnvVars("padding ${VARNAME} padding");
+    Assert.assertEquals("padding varvalue padding", resolved);
+  }
+
```
jholoman, line 48: Can you add a test for multiple substitutions per key?

flume-ng-doc/sphinx/FlumeUserGuide.rst
bessbd: I totally agree with 1&2. Thanks for those.
I'm pushing a commit soon to address the issues you have mentioned.

Regarding the "we" concern: we have a lot of "we"-s (sic :) ) in the doc: https://github.com/apache/flume/blob/trunk/flume-ng-doc/sphinx/FlumeUserGuide.rst#a-simple-example
We can either stay consistent with it or rephrase those using passive voice, etc.

flume-ng-node/src/test/java/org/apache/flume/node/TestEnvVarResolverProperties.java
bessbd: Just pushed a commit for this.
Please, let me know your thoughts.

meou:
LGTM. Thanks.

jholoman (approved):
LGTM +1

bessbd:
I'm about to commit this

bessbd:
@jholoman , @meou : thank you for the reviews!

PR #100, opened 2016-12-26
FLUME-2984: Fix OOM in thrift source. Make transport buffer configurable

laxman-ch:


laxman-ch:
Will close and reopen new pull request will right jira id so that it will be associated with appropriate jira.

PR #101, opened 2016-12-26
FLUME-2894: Fix flume components shutdown order

laxman-ch:


bessbd:
@laxman-ch : thank you for the patch!

It looks fine to me. However, to be able to commit it with confidence, some kind of test would be very useful.

It doesn't have to be a junit test, but any ways of making sure that this is the right change.

Ie: 
> Before applying the patch, if I do this, this happens.
> After applying the patch, if I do this, that happens.

Do you think you could please provide such additional information?

laxman-ch:
Thanks for the review @bessbd 

Before applying the patch that shutdown order is unpredictable and it is causing some data loss in our scenarios (AvroSource, CustomChannel, Kafka Sink)

After applying the patch that shutdown order is fixed (Sources, Sinks,  Channels) and the shutdown issues are solved (hangs, data losses)

I found it difficult to add a test for the shutdown scenario. Will check if its feasible.

PR #102, opened 2016-12-26
FLUME-2984: Fix OOM in thrift source. Make transport buffer configurable.

laxman-ch:


simonati (commented):
Looks good to me. 

Test passes. (mvn test -pl flume-ng-core -Dtest=org.apache.flume.source.TestThriftSource
) Unfortunately the test for the whole module fails for me with - I think irrelevant - missing snappy dependency.

simonati (commented):
Transport buffer safetynet was introduced when server was created by the getTThreadedSelectorServer(). Isn't something similar needed when the server creation felt back to getTThreadPoolServer()? Or the issue doesn't persists there?

bessbd:
@laxman-ch : thank you for the PR!

Is there a way to test this change? (If a unit test is difficult to implement, do you know of any ways to reproduce the issue manually? Ie. "this is the config I use, this is the input that causes OOM".)

asfgit:

PR #103, opened 2017-01-02
Add BucketWriterOptions to simplify the BucketWriter's constructor

adenes:
The goal if this pull request is to get rid of the 24 parameters of the `BucketWriter`'s constructor by wrapping them to the builder-like `BucketWriterOptions` class.


asfgit:

PR #104, opened 2017-01-11
fix "eventCounter" increase bug in append()

hangc0276:
In append() function, when the "batchCounter" is equal to batchSize, flush() will be called, which may throw exceptions when the network jitter happens on HDFS. Once exception happens, the event appended to the batch will be rollback, but the "eventCounter" has already been increased, which will make the "eventCounter" wrong value. So, the "eventCounter" should be increased after flush() happens.


asfgit:

PR #105, opened 2017-01-23
FLUME-3046: Kafka Sink and Source Configuration Improvements

tmgstevens:
Currently the Kafka Source sets the header for the topic. The sink reads this value, rather than the statically defined topic value. We should fix this so that you can either change the topic header that is used, or just choose to prefer the statically defined topic in the sink.

Adds configuration to Kafka Source to either disable the "topic" header, or at least configure it, and the same for the Kafka Sink,


flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSinkConstants.java
```diff
@@ -33,7 +33,10 @@ Licensed to the Apache Software Foundation (ASF) under one or more
       KAFKA_PREFIX + CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG;
 
   public static final String KEY_HEADER = "key";
-  public static final String TOPIC_HEADER = "topic";
+  public static final String TOPIC_DEFAULT_HEADER_NAME = "topic";
```
simonati: Please rename it to DEFAULT_<< property name >> to follow convention.

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
```diff
@@ -171,8 +173,8 @@ public Status process() throws EventDeliveryException {
         byte[] eventBody = event.getBody();
         Map<String, String> headers = event.getHeaders();
 
-        eventTopic = headers.get(TOPIC_HEADER);
-        if (eventTopic == null) {
+        eventTopic = headers.get(topicHeader);
```
simonati: Would you mind rephrasing it to 
```java
if (allowTopicHeader) {
    eventTopic = headers.get(topicHeader);  
    if (eventTopic == null) {
       log.debug("usage of topicheader was configured but header was not available {}", info about the event); 
       eventTopic = topic;
    }
} else {
    eventTopic = topic;
}
```

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
```diff
@@ -116,6 +115,9 @@ Licensed to the Apache Software Foundation (ASF) under one or more
   private boolean useAvroEventFormat;
   private String partitionHeader = null;
   private Integer staticPartitionId = null;
+  private boolean allowTopicHeader;
```
simonati: I think allowTopicOverride would be more self describing name.

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSinkConstants.java
```diff
@@ -33,7 +33,10 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
+  public static final String TOPIC_HEADER_NAME = "topicHeader";
+  public static final String ALLOW_TOPIC_HEADER = "allowTopicHeader";
```
simonati: please consider my naming comment from above (ALLOW_TOPIC_OVERRIDE)

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSinkConstants.java
simonati: Also I think TOPIC_OVERRIDE_HEADER and DEFAULT_TOPIC_OVERRIDE_HEADER naming might be better. Would you consider changing it?

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
```diff
@@ -49,9 +49,13 @@
   public static final String OLD_GROUP_ID = "groupId";
 
   // flume event headers
-  public static final String TOPIC_HEADER = "topic";
+  public static final String TOPIC_HEADER_DEFAULT = "topic";
```
simonati: Please use naming convention for default values by prefixing DEFAULT in the name instead of postfixing: DEFAULT_TOPIC_HEADER

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
```diff
@@ -49,9 +49,13 @@
...
   public static final String KEY_HEADER = "key";
   public static final String TIMESTAMP_HEADER = "timestamp";
   public static final String PARTITION_HEADER = "partition";
 
+  public static final String SET_TOPIC_HEADER = "setTopicHeader";
+  public static final boolean SET_TOPIC_HEADER_DEFAULT = true;
```
simonati: DEFAULT_SET_TOPIC_HEADER like above

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
```diff
@@ -49,9 +49,13 @@
...
+  public static final String TOPIC_HEADER_CONF = "topicHeader";
```
simonati: You may simply use TOPIC_HEADER instead.

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
```diff
@@ -655,6 +655,76 @@ public void testDefaultSettingsOnReConfigure() throws Exception {
...
+   * @throws InterruptedException
+   * @throws EventDeliveryException
+   */
+  @Test
+  public void testTopicHeaderSetNotSet() throws InterruptedException, EventDeliveryException {
+    context.put(TOPICS, topic0);
+    kafkaSource.configure(context);
+    kafkaSource.start();
+
+    Thread.sleep(500L);
```
simonati, line 670: What is the purpose of having Thread.sleep  between startup of source and produce? Is it really needed?

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
```diff
@@ -655,6 +655,76 @@ public void testDefaultSettingsOnReConfigure() throws Exception {
...
+
+    kafkaServer.produce(topic0, "", "hello, world");
+
+    Thread.sleep(500L);
```
simonati, line 674: What is the purpose of having Thread.sleep  between produce and process? Is it really needed?

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
```diff
@@ -655,6 +655,76 @@ public void testDefaultSettingsOnReConfigure() throws Exception {
...
   }
 
+  /**
+   * Tests the availability (or not) of the topic header in the output events,
+   * based on the configuration parameters added in FLUME-3046
+   * @throws InterruptedException
+   * @throws EventDeliveryException
+   */
+  @Test
+  public void testTopicHeaderSetNotSet() throws InterruptedException, EventDeliveryException {
```
simonati: Would you mind splitting it up to 3 tests which are indeed test slightly different things?

1. header override was not configured --> expect that there is a header called `topic` and its value is `test0`
2. header override was not configured only the header name --> expect that newly created header is `customTopicHeader`<->`test0`
3. header override is disabled --> expect `topic`<->`test0` and `customTopicHeader`<->`null`

simonati (changes requested):
Overall it looks good, thanks for catching this issue (loop bug)!

I mostly have variable naming change requests only.

After applying the patch:
- junit tests all passed (mvn test -fae -pl flume-ng-sinks/flume-ng-kafka-sink,flume-ng-sources/flume-kafka-source) 
 - also site build created an error free flumeuserguide. yaay!!!

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
tmgstevens: This gives time for Kafka to process the message - avoids flakiness. There is precedent in other tests for this.

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
tmgstevens: This gives time for Kafka to process the message - avoids flakiness. There is precedent in other tests for this.

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
tmgstevens: Done

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSinkConstants.java
tmgstevens: Done

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSinkConstants.java
tmgstevens: Done

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
tmgstevens: Done

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
tmgstevens: I think that's a different setting

flume-ng-sinks/flume-ng-kafka-sink/src/test/java/org/apache/flume/sink/kafka/TestKafkaSink.java
```diff
@@ -293,9 +293,9 @@ public void testTopicFromConfHeader() throws UnsupportedEncodingException {
   public void testTopicNotFromConfHeader() throws UnsupportedEncodingException {
     Sink kafkaSink = new KafkaSink();
     Context context = prepareDefaultContext();
-    context.put(KafkaSinkConstants.ALLOW_TOPIC_HEADER, "false");
-    context.put(KafkaSinkConstants.TOPIC_HEADER_NAME, "foo");
-    context.put(KafkaSinkConstants.TOPIC_HEADER_NAME, "bar");
+    context.put(KafkaSinkConstants.ALLOW_TOPIC_OVERRIDE_HEADER, "false");
+    context.put(KafkaSinkConstants.TOPIC_OVERRIDE_HEADER, "foo");
+    context.put(KafkaSinkConstants.TOPIC_OVERRIDE_HEADER, "bar");
```
simonati: nit: this line of code is not needed

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
```diff
@@ -173,10 +173,18 @@ public Status process() throws EventDeliveryException {
         byte[] eventBody = event.getBody();
         Map<String, String> headers = event.getHeaders();
 
-        eventTopic = headers.get(topicHeader);
-        if (eventTopic == null || !allowTopicHeader) {
+        if (allowTopicOverride) {
+          eventTopic = headers.get(topicHeader);
+          if (eventTopic == null) {
+            logger.debug("{} was set to true but header {} was null. Producing to {}" +
```
simonati: Thanks for the updated log message. I hope this will help troubleshooters.

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
simonati: As discussed offline an async call is involved so keeping this.

flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka/TestKafkaSource.java
simonati: As discussed offline an async call is involved so keeping this.

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
```diff
@@ -49,9 +49,13 @@
...
   // flume event headers
-  public static final String TOPIC_HEADER = "topic";
+  public static final String DEFAULT_TOPIC_HEADER = "topic";
   public static final String KEY_HEADER = "key";
   public static final String TIMESTAMP_HEADER = "timestamp";
   public static final String PARTITION_HEADER = "partition";
 
+  public static final String SET_TOPIC_HEADER = "setTopicHeader";
+  public static final boolean DEFAULT_SET_TOPIC_HEADER = true;
+  public static final String TOPIC_HEADER_CONF = "topicHeader";
```
simonati: Please remove the _CONF postfix.

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
simonati: Isn't it similar concept to the one in Sink? If not then would you mind updating the name of the DEFAULT constant to be consistent? 
(I see that KEY_HEADER looks confusing then but I guess that can be fixed in a followup commit)

simonati (commented):
Only a single naming question related to the TOPIC_HEADER_CONF

flume-ng-sinks/flume-ng-kafka-sink/src/test/java/org/apache/flume/sink/kafka/TestKafkaSink.java
tmgstevens: Fixed

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
tmgstevens: Fixed.

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
tmgstevens: Done

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSourceConstants.java
tmgstevens: Done

simonati:
Hi @tmgstevens,

Could you please retrigger TravisCI (eg change in a commit message)? It failed because of an unrelated issue which has been fixed since. 

> [ERROR] Failure executing javac, but could not parse the error:
> [ERROR] javac: invalid target release: 1.8

simonati (approved):
+1
Although with this patch preserves the erroneous behaviour in the edge-case when both a sink and source is used with the default settings it didn't break backward compatibility (which I think is more important for the non edge cases. To change this behaviour it is enough to default allowTopicOverride in KafkaSink to false).  Also it is now clear from the docs what to do to break the infinite cycle. Junit tests were passed. Time for test execution didn't increase materially so it is ok. For TestKafkaSource it took 2:07min vs 1:49min and for TestKafkaSink 42s vs 41s to pass)

Commands used for testing:
mvn test -pl flume-ng-sources/flume-kafka-source
mvn test -pl flume-ng-sinks/flume-ng-kafka-sink

PR #106, opened 2017-01-27
FLUME-3049: Remove exception wrapping in UGIExecutor.execute()

adenes:


simonati (approved):
Wrapping IOException into something else makes it harder for the caller to decide what to do next. Actually it hides the original information.

Checked that org.apache.flume.auth.SecurityException  is not used within flume codebase in execution control statements (in fact it is only being thrown).  Also the other implementation of FlumeAuthenticator (SimpleAuthenticator) doesn't do any wrapping like this so this behaviour seems inconsistent.

Classes in auth module is something that plugin dev can depend on but this change won't affect the public API, it would introduce only a behavioural change which I think acceptable in this situation since it fixes an undesired side effect.

Tests were run 
mvn clean install -DskipTests; mvn test -fae
and this time it all passed for me on OSX, java7. (yaay!!!)

mpercy:
Is there some way to write a test for this?

adenes:
Hi @mpercy, I've added two test cases to test whether the exception gets properly propagated in the `SimpleAuthenticator` and `KerberosAuthenticator` implementations. Let me know your opinion about these tests. Thanks.

flume-ng-auth/src/test/java/org/apache/flume/auth/TestFlumeAuthenticator.java
```diff
@@ -110,6 +111,46 @@ public void testFlumeLogin() throws IOException {
...
+   * Test whether the exception raised in the <code>PrivilegedExceptionAction</code> gets
+   * propagated as-is from {@link KerberosAuthenticator#execute(PrivilegedExceptionAction)}.
+   */
+  @Test(expected = IOException.class)
+  public void testKerberosAuthenticatorExceptionInExecute() throws Exception {
+    String principal = flumePrincipal;
+    String keytab = flumeKeytab.getAbsolutePath();
+
+    // Clear the previous statically stored logged in credentials
+    FlumeAuthenticationUtil.clearCredentials();
```
simonati: I think this cleanup should be part of tearDown. (Here and in other tests as well.)

simonati (changes requested):
The testing code requires some attention.

flume-ng-auth/src/test/java/org/apache/flume/auth/TestFlumeAuthenticator.java
adenes: Thanks @simonati for the comment, I pushed a new commit to refactor this.

simonati (approved):
LGTM+1
junit test passes for the new code as well:"mvn test -pl flume-ng-auth"

mpercy:
Your new test validates this change at a very low level, but doesn't actually replicate any aspect of the real environment. Can you think of a way to test the actual fix at a higher level?

More concretely: By making this change I think you are trying to fix a bug where the files never roll. But I don't see that issue filed or documented anywhere. I also don't see a test that shows that bug now being fixed. Do you think it's possible to write a test demonstrating that your patch fixes the bug?

Also, the commit message doesn't contain any information about why this change was necessary. From looking at FLUME-3049 I see a comment saying "This has unexpected side effects as the callers try to catch the wrapped exception". What are the side effects? What is the contract we are trying to provide to callers and how does this restore that? A more detailed explanation would be helpful for reviewers now and for developers / users in the future.

adenes:
Hi @mpercy, thanks for your comment. Yes, one of the side effects of the wrapping is that `BucketWriter` doesn't close the file if an `IOException` is thrown because it doesn't expect the exception to be wrapped in `SecurityException`. But it's just one example, same applies to the `BucketWriter.doFlush()` method, which is transitively called from `HDFSEventSink.process()` which also handles `IOException`s differently than other exceptions. I agree that this test is a bit low-level but I think writing unit test for one of the actual effects of the wrapping wouldn't really test this change.
My intent with the test was that the implementations (`SimpleAuthenticator` and `KerberosAuthenticator` - which delegates the call to `UGIExecutor`) don't wrap the exception, meaning they rethrow it as-is or a new one with the same type (or subclass). This is the contract we broke with the wrapping in the `UGIExecutor.execute()`. (I might have used other approaches to test, e.g. checking that the very same exception instance is thrown, that's true.)

I updated the Jira ticket to make the intent behind this change more clear.

mpercy:
> I think writing unit test for one of the actual effects of the wrapping wouldn't really test this change.

Why not?

mpercy:
Maybe it's not possible to test this change without running a MiniKdc or something. I haven't looked into it. But if it's possible to add a better test for this bug I think we should do it.

adenes:
I might be wrong but I think because the closing of the files depends on a couple of other components as well, I definitely wouldn't call that kind of a test a unit test for this change.
This test wants to test only the gist of this change: that the `PrivilegedExecutor` implementations propagate the original exceptions. One of the side effects of not doing so is that `BucketWriter` doesn't close the files (that was the original problem which led me to this bug), but there are other places where we call `PrivilegedExecutor.execute()`. None of those are prepared for `SecurityException`, so if we wanted to test this from the callers point of view we would need to create test cases for all the calls. 

I'm not against doing it, just my opinion is that they are simply too far from the change, the resulting test would be way too complex and hard to overview and would test many other things too. I would even call them integration tests, as they test a whole setup: Flume with some source and channel + hdfs sink with kerberos.

Again, I might be wrong and it may be possible to write better unit test for this but I couldn't come up with any significant improvement without making it unnecessary complex. If you have any concrete idea please share with me.

mpercy:
How about adding a test to TestBucketWriter that uses a UGIExecutor? You can modify or subclass MockHDFSWriter to cause an exception to be thrown on the following call to open() or flush(), for example. Then write a test to ensure that rolling occurs in the situation where it wouldn't before.

Even better, while we're doing this we could improve our test coverage of BucketWriter in a secured environment in general. I don't think we have any unit test coverage of that right now, which is unfortunate (am I wrong about that?). You could add a test for this case without specifying which PrivilegedExecutor to use (i.e. just use the proxy object that is a protected member of TestBucketWriter), then parameterize TestBucketWriter so that we execute all test cases in both secure and non-secure modes.

We could test in a Kerberized env by instantiating a MiniKdc like we do in TestFlumeAuthenticator and using UGIExecutor with the same keytab we use there. Actually, I am not sure we even need Kerberos to use the UGIExecutor, since doAs isn't a Kerberos concept (it's a feature of HDFS).

Sound doable?

mpercy:
I haven't tested this, and probably we could do something even better, but maybe something similar to this would give us a reasonable amount of coverage for this case without a lot of boilerplate code or test setup needed. I'm assuming we already have unit test coverage for what to do when the bucketWriter append() call throws an IOException or when the bucketWriter is closed when HDFSEventSink tries to append an event to it:

```java
  // Test that we don't swallow IOExceptions in secure mode. We should close the bucket writer
  // and rethrow the exception. Regression test for KUDU-3049.
  @Test
  public void testRotateBucketOnIOException() throws IOException, InterruptedException {
    int maxBytes = 300;
    MockHDFSWriter hdfsWriter = new MockHDFSWriter();
    final String PROXY_USER = "alice";
    PrivilegedExecutor ugiProxy =
        FlumeAuthenticationUtil.getAuthenticator(null, null).proxyAs(PROXY_USER);
    BucketWriter bucketWriter = new BucketWriter(
        0, maxBytes, 0, 0, ctx, "/tmp", "file", "", ".tmp", null, null,
        SequenceFile.CompressionType.NONE, hdfsWriter, timedRollerPool, ugiProxy,
        new SinkCounter("test-bucket-writer-" + System.currentTimeMillis()), 0, null, null, 30000,
        Executors.newSingleThreadExecutor(), 0, 0);

    Event e = EventBuilder.withBody("foo", Charsets.UTF_8);
    // Write one event successfully.
    bucketWriter.append(e);
    
    // Fail the next write.
    // Note: We might be able to do this more generically using mockito.
    hdfsWriter.throwExceptionOnNextAppend(new IOException("Test injected IOException"));
    
    // The second time we try to write we should get an IOException.
    try {
      bucketWriter.append(e);
      fail();
    } catch (IOException ex) {
      logger.info("Caught expected IOException", ex);
    }
    // The third time we try to write we should get a BucketClosedException, because the
    // BucketWriter should attempt to close itself before rethrowing the IOException on the first
    // call.
    try {
      bucketWriter.append(e);
      fail();
    } catch (BucketClosedException ex) {
      logger.info("Caught expected BucketClosedException", ex);
    }

    assertEquals("events written", 1, hdfsWriter.getEventsWritten());
  }
```

adenes:
Thank you for the detailed explanation @mpercy. I incorporated your code snippet into `TestBucketWriter` with some minor changes (e.g. using Mockito as suggested).
I like the idea of changing it to parameterized test, I'll created a separate PR for that.

mpercy:
Thanks for updating the test, @adenes. +1, LGTM. I am about to commit this.

PR #108, opened 2017-02-20
FLUME-3056 TestApplication hangs indefinitely

andrasbeni:
Skip configuration handling in Application while stop() is running

### DIFF
--- a/flume-ng-node/src/main/java/org/apache/flume/node/Application.java
+++ b/flume-ng-node/src/main/java/org/apache/flume/node/Application.java
@@ -33,7 +33,6 @@ import org.apache.commons.cli.ParseException;
 import org.apache.flume.Channel;
 import org.apache.flume.Constants;
 import org.apache.flume.Context;
-import org.apache.flume.FlumeException;
 import org.apache.flume.SinkRunner;
 import org.apache.flume.SourceRunner;
 import org.apache.flume.instrumentation.MonitorService;
@@ -68,7 +67,7 @@ public class Application {
   private final LifecycleSupervisor supervisor;
   private MaterializedConfiguration materializedConfiguration;
   private MonitorService monitorServer;
-  private volatile Lock lock = new ReentrantLock();
+  private final Lock lifecycleLock = new ReentrantLock();
 
   public Application() {
     this(new ArrayList<LifecycleAware>(0));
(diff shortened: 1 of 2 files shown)


flume-ng-node/src/main/java/org/apache/flume/node/Application.java
```diff
@@ -83,16 +84,25 @@ public synchronized void start() {
...
+    }
   }
 
   public synchronized void stop() {
+    stopping = true;
     supervisor.stop();
     if (monitorServer != null) {
       monitorServer.stop();
     }
+    stopping = false;
```
adenes: In general it's a good practice to wrap this to a `finally` block. In this particular case it mightn't be necessary as this is only called from the shutdown hook, so in fact it doesn't matter whether this is set to `false`. But in that case do we really need this line, wdyt?

adenes (changes requested):
Good catch, thanks for the fix. I had one minor comment, otherwise 👍

simonati (commented):
Hi @andrasbeni,

Could you please explain the issue in detail? What exactly is in deadlock and why this change prevents it to happen?
Just guessing do we suffer from the same issue during start?

andrasbeni:
Hi @simonati ,
Please see my comment in [jira](https://issues.apache.org/jira/browse/FLUME-3056?focusedCommentId=15879129&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15879129) . I hope it describes the problem.

flume-ng-node/src/main/java/org/apache/flume/node/Application.java
andrasbeni: Thanks for your remark. I'm afraid I will have to rework my change a bit, because this single boolean flag will not entirely prevent the issue.

flume-ng-node/src/main/java/org/apache/flume/node/Application.java
```diff
@@ -65,6 +68,7 @@
   private final LifecycleSupervisor supervisor;
   private MaterializedConfiguration materializedConfiguration;
   private MonitorService monitorServer;
+  private volatile Lock lock = new ReentrantLock();
```
adenes: I'd choose a more descriptive name for this, e.g. `lifecycleLock`. I suppose the `volatile` just remained from the previous change, a `final` would be better here :)

adenes (approved):
Thank you for the patch @andrasbeni, it looks good to me, 👍

andrasbeni:
Retest this please

szaboferee (approved):
Tested it with putting a sleep in the file watcher to make sure that the application object is locked while shutdown.
Without the fix, the agent hangs, with the fix it stops successfully.

PR #98, opened 2016-12-12
Change how we build and sign release artifacts

mpercy:
These simple scripts automate creation and signing of the release
artifacts. They help guarantee that the source artifacts published match
the Git tag.

Updated the How To Release documentation to explain how to use the
scripts.

PR #109, opened 2017-02-24
flume-checkstyle module should use ASF root pom as parent

mpercy:
The flume-checkstyle module doesn't currently declare a parent pom. This
will be a problem when we try to deploy our artifacts because we are
missing a distributionManagement section in the flume-checkstyle module.
The deploy will fail with an error message that looks something like:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy)
on project flume-checkstyle: Deployment failed: repository element was not specified in the POM inside
distributionManagement element or in -DaltDeploymentRepository=id::layout::url parameter

This patch should solve the problem.

### DIFF
--- a/flume-checkstyle/pom.xml
+++ b/flume-checkstyle/pom.xml
@@ -20,7 +20,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-
+  <parent>
+    <groupId>org.apache</groupId>
+    <artifactId>apache</artifactId>
+    <version>18</version>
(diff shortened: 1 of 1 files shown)


adenes:
Thanks @mpercy for the fix. I'm wondering how this didn't cause a problem during the 1.7 release, do you have any idea, do you remember any issue around this?

mpercy:
We had an unrelated issue that caused us to back out the checkstyle patch for 1.7. The problem was that the source artifact wouldn't build with checkstyle. That was fixed in #98.

simonati (approved):
Change looks good to me.

adenes:
Thanks @mpercy for the clarification.
We've tested the change with a different parent pom (thus different repository settings), without the parent definition in place `mvn deploy` failed indeed, with the change it was successful.
👍  for the PR.

bessbd:
LGTM, +1
@mpercy : would you like me to commit this for you?

mpercy:
Thx for the +1 Donat! I can commit it, no worries.

PR #111, opened 2017-03-01
NPE fix

K0zka:
https://issues.apache.org/jira/browse/FLUME-3064

### DIFF
--- a/flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/ChannelCallbackTest.java
+++ b/flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/ChannelCallbackTest.java
@@ -1,0 +1,66 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flume.channel.kafka;
+
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ChannelCallbackTest {
+
+  //note that this is log4j backing the slf4j logger
+  //slf4j does not give an API to change log level at runtime
+  private static final Logger CHANNEL_BACKING_LOGGER = Logger.getLogger(ChannelCallback.class);
+
+  Level originalLevel;
+
+  @Before
+  public void setup() {
+    originalLevel = CHANNEL_BACKING_LOGGER.getLevel();
+  }
+
+  @After
+  public void cleanup() {
+    CHANNEL_BACKING_LOGGER.setLevel(originalLevel);
+  }
+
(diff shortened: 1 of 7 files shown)


adenes (commented):
I think the "Acked message..." should be logged only if the record was successfully sent (= there was no exception).
So I'd add an `else` branch to the `if (exception != null)` which would implicitly make sure that the `metadata` is not `null`.
(As per the `Callback`'s javadoc: "Exactly one of the arguments will be non-null.")
What do you think?

flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/ChannelCallbackTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ChannelCallbackTest {
```
adenes, line 27: Could you please rename this to `TestChannelCallback` to conform to the current naming conventions?

flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/ChannelCallbackTest.java
```diff
@@ -0,0 +1,66 @@
...
+    CHANNEL_BACKING_LOGGER.setLevel(originalLevel);
+  }
+
+  private final ChannelCallback channelCallback = new ChannelCallback(0, System.currentTimeMillis());
+  private final RecordMetadata metadata
+          = new RecordMetadata(new TopicPartition("TEST-TOPIC", 0), 0, 0);
+  private final Exception exception = new Exception("TEST");
+
+  @Test
+  public void onCompletion() throws Exception {
```
adenes, line 51: Could you please add some comment here, maybe with a reference to the Jira issue? Knowing the background helps to understand the test and that the expected behaviour is not to throw an exception but without that knowledge it's quite baffling.
It would be even better to wrap the `onCompletion` calls to a `try-catch` block and in case of an (unwanted) NPEx fail the test with an `Assert.fail()`.

Please also rename the test method name to `testOnCompletion` as most of the test methods in Flume have the `test` prefix. (Or to some more self-explaining like `testOnCompletionLoggingNoException` - could be better but this is the best I could come up with)

flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/ChannelCallbackTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+public class ChannelCallbackTest {
+
+  //note that this is log4j backing the slf4j logger
+  //slf4j does not give an API to change log level at runtime
+  private static final Logger CHANNEL_BACKING_LOGGER = Logger.getLogger(ChannelCallback.class);
+
+  Level originalLevel;
```
adenes, line 33: minor: this can be private

flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/ChannelCallbackTest.java
```diff
@@ -0,0 +1,66 @@
...
+  public void setup() {
+    originalLevel = CHANNEL_BACKING_LOGGER.getLevel();
+  }
+
+  @After
+  public void cleanup() {
+    CHANNEL_BACKING_LOGGER.setLevel(originalLevel);
+  }
+
+  private final ChannelCallback channelCallback = new ChannelCallback(0, System.currentTimeMillis());
```
adenes, line 45: minor: could you please move these 3 variables up to the top of the class?

adenes (changes requested):
I added a couple of comments to the `ChannelCallbackTest` but those apply to the `SinkCallbackTest` as well.

Beside those there is one checkstyle violation, the ChannelCallbackTest:46 line is longer than 100 characters. You can verify whether your code conforms to the checkstyle rules by running `mvn verify -DskipTests`.

K0zka:
no, no fucking way :-D I have wasted way too much time on this unfortunately, I'd rather just remember not to set the log level to debug

adenes:
ok, no problem, I'm going to resubmit this if you have no objections.

PR #112, opened 2017-03-07
"[FLUME-3067][Shell]Defalut JAVA_OPTS "-Xmx20m" may be conflicted with cuctom defines."

chenjianwei3:
When write a Flume bootstrap shell, set the environment variable FLUME_JAVA_OPTS with the defined values as the min(-Xms) 
and max(-Xmx) heap size for the JVM, then call "bin/flime-ng" to start Flume. 
Succeed to start, but the defalut "-Xmx20m" is still in process JVM options.

### DIFF
--- a/bin/flume-ng
+++ b/bin/flume-ng
@@ -227,7 +227,7 @@ run_flume() {
 FLUME_CLASSPATH=""
 FLUME_JAVA_LIBRARY_PATH=""
 JAVA_OPTS=""
-JAVA_HEAP_MAX ="-Xmx20m"
+JAVA_HEAP_MAX="-Xmx20m"
 LD_LIBRARY_PATH=""
 
 opt_conf=""
--- a/bin/flume-ng
+++ b/bin/flume-ng
@@ -212,7 +212,7 @@ run_flume() {
     set -x
   fi
   
-  if [[ $FLUME_JAVA_OPTS =~ "Xmx" ]] || [[ $JAVA_OPTS =~ "Xmx" ]]; then
+  if [[ $FLUME_JAVA_OPTS =~ Xmx[0-9]+ ]] || [[ $JAVA_OPTS =~ Xmx[0-9]+ ]]; then
     unset JAVA_HEAP_MAX 
   fi
   $EXEC $JAVA_HOME/bin/java $JAVA_HEAP_MAX $JAVA_OPTS $FLUME_JAVA_OPTS "${arr_java_props[@]}" -cp "$FLUME_CLASSPATH" \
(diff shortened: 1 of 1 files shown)


bin/flume-ng
```diff
@@ -211,6 +211,10 @@ run_flume() {
   if [ ${CLEAN_FLAG} -ne 0 ]; then
     set -x
   fi
+  
+  if [[ $FLUME_JAVA_OPTS =~ "Xmx" ]]; then
```
adenes: I think it might be good to rename the `$JAVA_OPTS` to `$JAVA_HEAP_MAX` or add this as a new variable (as in Hadoop: https://github.com/apache/hadoop-common/blob/trunk/hadoop-common-project/hadoop-common/src/main/bin/hadoop-functions.sh#L544).
The reason is that if in the future new default parameters will be added to the `$JAVA_OPTS` passing in `-Xmx` will clear them too, resulting in unexpected behaviour (unless we do this change when we add the new parameters).

adenes (changes requested):
Thank you for the patch. I added one comment otherwise looks good to me, 👍

bin/flume-ng
```diff
@@ -222,7 +226,8 @@ run_flume() {
 # set default params
 FLUME_CLASSPATH=""
 FLUME_JAVA_LIBRARY_PATH=""
-JAVA_OPTS="-Xmx20m"
+JAVA_OPTS=""
+JAVA_HEAP_MAX ="-Xmx20m"
```
adenes: Please remove the extra space before the `=`, it results in runtime error (`bin/flume-ng: line 230: JAVA_HEAP_MAX: command not found`). Thank you.

chenjianwei3: so careless,   thank you very much.

bin/flume-ng
```diff
@@ -211,7 +211,11 @@ run_flume() {
   if [ ${CLEAN_FLAG} -ne 0 ]; then
     set -x
   fi
-  $EXEC $JAVA_HOME/bin/java $JAVA_OPTS $FLUME_JAVA_OPTS "${arr_java_props[@]}" -cp "$FLUME_CLASSPATH" \
+  
+  if [[ $FLUME_JAVA_OPTS =~ "Xmx" ]] || [[ $JAVA_OPTS =~ "Xmx" ]]; then
```
adenes: We could make it more bulletproof by using `$FLUME_JAVA_OPTS =~ -Xmx[0-9]+` check to make sure that there is a dash before and at least one number after the `Xmx`.
Not sure that this is necessary though, what do you think, @chenjianwei3 ?

adenes (approved):
Thank you for the patch and being responsive during the review. LGTM, 👍

asfgit:

PR #113, opened 2017-03-09
[FLUME-3068] Ignore hidden files in TaildirMatcher

lingjinjiang:
[https://issues.apache.org/jira/browse/FLUME-3068](url)
When using the taildir to monitor a directory, hidden files in the directory are tailed by the source.
As the hidden files are often temp files, config files or the files which not want to be tailed, so it's necessary to ignore the hidden file when using the taildir source.


adenes:
Hi @lingjinjiang,
Thank you for the patch. My only concern with it is that with this we won't be backward compatible, although I'm not totally sure that this is necessary in this case.
On the other hand it wouldn't need significant effort to introduce a new configuration option for this (which by default would allow listing the hidden files).
What do you (and the others) think?

lingjinjiang:
Hi @adenes ,
Thank you for your advice. Now I add a new configuration option named "ignoreHiddenFile", and the default value is "false". When set it to "true", the hidden file will be ingored.

adenes:
Hi @lingjinjiang, thank you for the update on the patch and sorry for the delayed reply.
Could you please also add unit tests for the feature and update the documentation as well? Thank you.

lingjinjiang:
Hi @adenes , I add the unit test and update the documents. Could you review it?

asfgit:

PR #114, opened 2017-03-09
"idleFuture" should reset to "null" after idle bucket closed

hangc0276:
If we use "idleTimeout" to roll file, once the flush() method called, it will start a delayed thread to close the idle bucketWriter. The delayed thread will call close() method to check whether the "idleFuture" is done, and close opened file. If the "idleFuture" is running when checking, the idleFuture.isDone() method will return false, and the idleFuture won't be set to "null". When the bucketWriter calls flush() on the next time, the "idleFuture" is not "null" and idleFutrue.cancel(false) method returns false due to the "idleFuture" has completed in the last call. So new thread won't be created to close current idle bucketWriter and keeps current file openning.

### DIFF
--- a/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
@@ -440,6 +440,7 @@ class BucketWriter {
               if (isOpen) {
                 close(true);
               }
+              idleFuture = null;
               return null;
             }
           };


adenes:
Hi @hangc0276,
Thank you for your patch.

Unless I misunderstood your explanation I think it can't happen. Let me summarize it, and please correct me if I got it wrong:
- `flush()` schedules the `idleFuture`
- in `close()` if `idleFuture != null && !idleFuture.isDone()` -> cancelling `idleFuture` and setting to `null`
But this is a _not done_ check, so if it's not done yet, it will be set to `null` (regardless if it's currently running, which means that the `close()` was called by it)

In fact the `idleFuture` will be set to `null` for sure:
- if in the previously mentioned check `idleFuture.isDone()` is `true` that would mean that it had been run before -> called the `close()` method -> set the `BucketWriter.idleFuture` to `null`.
- if it's not done yet -> the `BucketWriter.idleFuture` will be set to `null`

Let me know if you see any flaw in my reasoning.

asfgit:

PR #115, opened 2017-03-09
Doc fix: Size of files created by HDFS Sink won't necessarily match the configured rollSize value

adenes:
Updated the description of the HDFS Sink's `hdfs.rollSize` property to reflect the actual behaviour: it keeps track of the size of the incoming events' body thus the size of the created files won't necessarily match the configured value.


liorze:
:ship:

marcellhegedus (approved):

asfgit:

PR #120, opened 2017-03-14
FLUME-2579 Support durable subscription in JMSSource

andrasbeni:
JMSSource has created only nondurable subscriptions which could lead to event loss in case of topic destination type.

This change enables durable subscription creation and lets user specify client id.
Also removed JMSMessageConsumerFactory which has no additional value.


flume-ng-sources/flume-jms-source/src/main/java/org/apache/flume/source/jms/JMSMessageConsumerFactory.java
```diff
@@ -1,36 +0,0 @@
...
- * limitations under the License.
- */
-package org.apache.flume.source.jms;
-
-import javax.jms.ConnectionFactory;
-import javax.naming.InitialContext;
-
-import com.google.common.base.Optional;
-
-public class JMSMessageConsumerFactory {
```
simonati, line 25: Thanks for cleaning this up.

simonati (approved):
LGTM

flume-ng-sources/flume-jms-source/src/main/java/org/apache/flume/source/jms/JMSSource.java
```diff
@@ -121,6 +121,13 @@ protected void doConfigure(Context context) throws FlumeException {
     pollTimeout = context.getLong(JMSSourceConfiguration.POLL_TIMEOUT,
         JMSSourceConfiguration.POLL_TIMEOUT_DEFAULT);
 
+    clientId = Optional.fromNullable(context.getString(JMSSourceConfiguration.CLIENT_ID));
+
+    createDurableSubscription = context.getBoolean(
+        JMSSourceConfiguration.CREATE_DURABLE_SUBSCRIPTION, false);
```
simonati: Would you mind using constants instead of magic value? ie DEFAULT_CREATE_DURABLE_SUBSCRIPTION and DEFAULT_DURABLE_SUBSCRIPTION_NAME

adenes (approved):
Thank you for the patch, LGTM, 👍

kailashk24:
Is this fix available as a patch ?

adenes:
@kailashk24 , it hasn't been merged to the main branch yet, but I'm going to commit it soon if there are no objections.

PR #121, opened 2017-03-15
FLUME-3072 Add IP address to headers in flume log4j appender

andrasbeni:
Log4jAppender and LoadBalancingLog4jAppender resolve local hosts address at startup.
The hostname is added to each event's header with the key "flume.client.log4j.address".

### DIFF
--- a/flume-ng-clients/flume-ng-log4jappender/src/main/java/org/apache/flume/clients/log4jappender/Log4jAppender.java
+++ b/flume-ng-clients/flume-ng-log4jappender/src/main/java/org/apache/flume/clients/log4jappender/Log4jAppender.java
@@ -20,6 +20,8 @@ package org.apache.flume.clients.log4jappender;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.net.InetAddress;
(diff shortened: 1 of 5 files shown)


flume-ng-clients/flume-ng-log4jappender/src/main/java/org/apache/flume/clients/log4jappender/Log4jAppender.java
```diff
@@ -135,7 +138,7 @@ public synchronized void append(LoggingEvent event) throws FlumeException {
     hdrs.put(Log4jAvroHeaders.LOGGER_NAME.toString(), event.getLoggerName());
     hdrs.put(Log4jAvroHeaders.TIMESTAMP.toString(),
         String.valueOf(event.timeStamp));
-
+    hdrs.put(Log4jAvroHeaders.ADDRESS.toString(), clientAddress);
```
simonati, line 141: I'm fine with adding an extra header to the events by default.

simonati (approved):
LGTM

adenes (approved):
Thank you for the patch, LGTM 👍

PR #123, opened 2017-03-20
Added support for Filter Queries read from config file

scottkp:
Added support for optional property in config file for Twitter Source. E.g. if keywords property is provided:
...
a1.sources.r1.type = org.apache.flume.source.twitter.TwitterSource
a1.sources.r1.keywords = word1,word2,word3
...

Otherwise, existing functionality of twitterStream.sample() is invoked.


asfgit:

PR #125, opened 2017-03-29
bugfix: when loglevel is debug, Kafka Sink will throw NPE because of …

loleek:
When we send a event with SinkCallback to kafka under DEBUG level of log4j, if kafka reponses a result with exception but not a RecordMetadata. Kafka Sink will throw NPE.
code in SinkCallback:
if (logger.isDebugEnabled()) {
  long eventElapsedTime = System.currentTimeMillis() - startTime;
  logger.debug("Acked message partition:{} ofset:{}",  metadata.partition(), metadata.offset());
  logger.debug("Elapsed time for send: {}", eventElapsedTime);
}
code in Kafka Producer:
if (exception == null) { 
  RecordMetadata metadata = new RecordMetadata(...);
  thunk.callback.onCompletion(metadata, null);
} else {
  thunk.callback.onCompletion(null, exception);
}

I've already created this issue on jira.
https://issues.apache.org/jira/browse/FLUME-3043?jql=project%20%3D%20FLUME

### DIFF
--- a/flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
+++ b/flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
@@ -453,7 +453,10 @@ class SinkCallback implements Callback {
 
     if (logger.isDebugEnabled()) {
       long eventElapsedTime = System.currentTimeMillis() - startTime;
-      logger.debug("Acked message partition:{} ofset:{}",  metadata.partition(), metadata.offset());
+      if (metadata != null) {
+        logger.debug("Acked message partition:{} ofset:{}", metadata.partition(),
+                metadata.offset());
+      }
       logger.debug("Elapsed time for send: {}", eventElapsedTime);
     }
   }
--- a/flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
+++ b/flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
@@ -453,7 +453,9 @@ class SinkCallback implements Callback {
 
     if (logger.isDebugEnabled()) {
       long eventElapsedTime = System.currentTimeMillis() - startTime;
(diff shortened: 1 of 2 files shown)


flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
```diff
@@ -453,7 +453,9 @@ public void onCompletion(RecordMetadata metadata, Exception exception) {
 
     if (logger.isDebugEnabled()) {
       long eventElapsedTime = System.currentTimeMillis() - startTime;
-      logger.debug("Acked message partition:{} ofset:{}",  metadata.partition(), metadata.offset());
+      if (metadata != null) {
+        logger.debug("Acked message partition:{} ofset:{}", metadata.partition(), metadata.offset());
```
adenes: This line is longer than 100 characters which brakes the checkstyle verification, could you please fix it? Thank you.

adenes (changes requested):
Hi @loleek, thank you for your patch.
`KafkaChannel` has the same logic, could you please change that as well? It would be great, thanks in advance.

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
loleek: My pleasure

loleek: Done.Thanks for you review.

loleek:
Hi @adenes , things already done. Thanks for your suggestion.

adenes (approved):
Thank you for the patch, it looks good to me, 👍

loleek:
@adenes Excause me, this is my first pull request.I wonder what's the next after your review.

adenes:
Unless an other reviewer finds anything to change in this PR there is nothing to do on your side. A committer will commit this soon. Thanks again for the fix.

K0zka:
This change is so simple, why does it take so long? It is sad to see how poorly this essential component is maintained :-(

bessbd:
@loleek : thank you for the patch!

bessbd:
@adenes : thank you for the review!

PR #127, opened 2017-04-03
FLUME-3080. Call DistributedFileSystem.recoverLease() if close() fails to avoid lease leak

adenes:
If the HDFS Sink tries to close a file but it fails (e.g. due to timeout) the last block might not end up in COMPLETE state. In this case block recovery should happen but as the lease is still held by Flume the NameNode will start the recovery process only after the hard limit of 1 hour expires.

This change adds an explicit recoverLease() call in case of close failure.

For more details see https://issues.apache.org/jira/browse/FLUME-3080

### DIFF
--- a/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
@@ -30,6 +30,7 @@ import org.apache.flume.sink.hdfs.HDFSEventSink.WriterCallback;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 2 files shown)


harishreedharan:
+1. LGTM. Please feel free to merge @adenes !

adenes:
Thank you for the review @harishreedharan. I'd be more than happy to commit but unfortunately I don't have committer rights yet.

bessbd:
@harishreedharan : I can commit it if you'd like me to.

adenes:
Hi @bessbd, it'd be great if you could commit this. Thanks in advance.

bessbd:
@adenes : I'll do it asap

bessbd:
@harishreedharan : thank you for the review!

bessbd:

adenes:
Thank you @bessbd for committing this. Could you please have a look on the other open PRs as well? Thanks.

bessbd:
@adenes : sure. I've just been a little swamped lately.
I've noticed that you have reviewed many of those open PRs. Do you mind providing single-line comments/summaries on future reviews? (Ie. impact of patch and/or things you have considered for approval)

PR #129, opened 2017-04-20
FLUME-3085: HDFS Sink can skip flushing some BucketWriters, might lead to data loss

adenes:
This commit fixes the issue when in `HDFSEventSink.process()` a `BucketWriter.append()` call threw a `BucketClosedException` then the newly created `BucketWriter` wasn't flushed after the processing loop.

### DIFF
--- a/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
@@ -22,8 +22,10 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.TimeZone;
 import java.util.Map.Entry;
 import java.util.concurrent.ExecutorService;
@@ -40,7 +42,6 @@ import org.apache.flume.EventDeliveryException;
 import org.apache.flume.SystemClock;
 import org.apache.flume.Transaction;
 import org.apache.flume.auth.FlumeAuthenticationUtil;
-import org.apache.flume.auth.FlumeAuthenticator;
 import org.apache.flume.auth.PrivilegedExecutor;
 import org.apache.flume.conf.Configurable;
 import org.apache.flume.formatter.output.BucketPath;
@@ -55,7 +56,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 public class HDFSEventSink extends AbstractSink implements Configurable {
@@ -354,9 +354,9 @@ public class HDFSEventSink extends AbstractSink implements Configurable {
   public Status process() throws EventDeliveryException {
     Channel channel = getChannel();
     Transaction transaction = channel.getTransaction();
-    List<BucketWriter> writers = Lists.newArrayList();
     transaction.begin();
     try {
+      Set<BucketWriter> writers = new LinkedHashSet<>();
       int txnEventCount = 0;
       for (txnEventCount = 0; txnEventCount < batchSize; txnEventCount++) {
         Event event = channel.take();
@@ -396,11 +396,6 @@ public class HDFSEventSink extends AbstractSink implements Configurable {
           }
         }
 
-        // track the buckets getting written in this transaction
-        if (!writers.contains(bucketWriter)) {
-          writers.add(bucketWriter);
-        }
-
         // Write the data to HDFS
         try {
           bucketWriter.append(event);
@@ -415,6 +410,9 @@ public class HDFSEventSink extends AbstractSink implements Configurable {
           }
           bucketWriter.append(event);
         }
+
+        // track the buckets getting written in this transaction
+        writers.add(bucketWriter);
       }
 
       if (txnEventCount == 0) {
@@ -455,7 +453,8 @@ public class HDFSEventSink extends AbstractSink implements Configurable {
     }
   }
 
-  private BucketWriter initializeBucketWriter(String realPath,
+  @VisibleForTesting
+  BucketWriter initializeBucketWriter(String realPath,
       String realName, String lookupPath, HDFSWriter hdfsWriter,
       WriterCallback closeCallback) {
     BucketWriter bucketWriter = new BucketWriter(rollInterval,
--- a/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
@@ -412,7 +412,9 @@ public class HDFSEventSink extends AbstractSink implements Configurable {
         }
 
         // track the buckets getting written in this transaction
-        writers.add(bucketWriter);
+        if (!writers.contains(bucketWriter)) {
+          writers.add(bucketWriter);
+        }
       }
 
       if (txnEventCount == 0) {
--- a/flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import org.apache.avro.file.DataFileStream;
 import org.apache.avro.generic.GenericData;
@@ -67,6 +68,9 @@ import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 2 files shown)


adenes:
Some comments for the implementation:
- I replaced the `writers` list to a set. Actually to a `LinkedHashSet` in order to be consistent with the `List` implementation regarding to the ordering of the elements. (Although the logic doesn't depend on the ordering.)
- The `writers` variable was moved into the `try-catch` block as it isn't used outside of it.
- I moved the `writers.add(bucketWriter)` call to right after the `append()` call. It means that the first `BucketWriter` won't be added if it has been closed already (so a new `BucketWriter` was created), but it isn't necessary as flushing a closed `BucketWriter` does nothing (and flush happened during the close, anyway).
- I had to change the visibility of the `initializeBucketWriter()` method to package private to be able to mock the `BucketWriter`s in the test.

Tests in `flume-ng-sinks/flume-hdfs-sink` pass, checkstyle as well.

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1545,4 +1549,61 @@ private void testRetryRename(boolean closeSucceed)
...
+    channel.put(EventBuilder.withBody("test".getBytes()));
+    txn.commit();
+    txn.close();
+
+    sink.process();
+
+    // After processing the events the only BucketWriter in the sfWriters map is
+    // the one which was created after the first one threw the BucketClosedException.
+    // It is expected that its flush() method was called exactly once.
+    Map<String, BucketWriter> bucketWriterMap = sink.getSfWriters();
```
simonati: Instead of accessing sfwriters you can collect the created BucketWriters from your test extension of HDFSEventSink. This way you could check that whatever bucketwriter was created is actually flushed/closed. sfwriters list might not have reference to every created bucketwriter (lookup path is common so the newly created bw kicks out the closed one on line 406).

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
```diff
@@ -415,6 +410,9 @@ public void run(String bucketPath) {
           }
           bucketWriter.append(event);
         }
+
+        // track the buckets getting written in this transaction
+        writers.add(bucketWriter);
```
simonati: Looks good to also capture the bucketwriter created on line 406. Since the original (line 389 or 393) one threw a closed exception that means that we don't have to flush the original one only the newly created one (close implicitly contains flush as well)

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1545,4 +1549,61 @@ private void testRetryRename(boolean closeSucceed)
...
+    Context context = new Context(ImmutableMap.of("hdfs.path", testPath));
+    Configurables.configure(sink, context);
+
+    MemoryChannel channel = new MemoryChannel();
+    Configurables.configure(channel, new Context());
+
+    sink.setChannel(channel);
+    sink.start();
+
+    Transaction txn = channel.getTransaction();
```
simonati: nit: I would have stubbed the channel.take() instead of juggling around with this transaction.

simonati (changes requested):
Business logic looks good to me. 
Improvement requested on testing part only as it test a slightly different thing (sfWriter list vs all bucketwriters).
mvn test -pl flume-ng-sinks/flume-hdfs-sink passes.

adenes:
Thank you for the review @simonati, I have updated the change based on your comments, please let me know what you think.

simonati:
Hi @bessbd, @mpercy,

Could you please retrigger Travis CI? (build failed because of https://repo.maven.apache.org/maven2 had been unreachable)
I doesn't seem to have the necessary entitlement to do it. 

Cheers,
Attila

mpercy:
@simonati I don't know how to do that; try just re-pushing the patch.

simonati:
Hi @mpercy,
For future reference a workaround exists: amending the commit message and force pushing the last commit retriggered Travis CI. (And any meaningful new commit would do the same)

adenes:
@mpercy, @bessbd, could you please review this change and commit if you are ok with it? Thanks in advance.

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1545,4 +1553,62 @@ private void testRetryRename(boolean closeSucceed)
...
+        }
+        bucketWriters.add(bw);
+        return bw;
+      }
+    };
+
+    Context context = new Context(ImmutableMap.of("hdfs.path", testPath));
+    Configurables.configure(sink, context);
+
+    Channel channel = Mockito.spy(new MemoryChannel());
```
simonati, line 1592: Ahh I see. MockChannel is in flume-ng-core test folder thus cannot be used as is. And a simple Mockito.mock(MemoryChannel.class) is not enough as beside take(), also getTransaction has to be implemented. It might need a later refactoring to move this Mock implementations to a shared module but for now I'm fine with this.

simonati (approved):
LGTM

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
mpercy: nit: Can we keep the if-statement? I suppose add() on the same element multiple times would just replace the previous element in the set and end up as a no-op, but it seems clearer to me to keep the "contains" there, especially now that you've switched to a hash-based container and it would be an O(1) operation, e.g.

```
if (!writers.contains(bucketWriter)) {		
  writers.add(bucketWriter);		
}
```

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
```diff
@@ -354,9 +354,9 @@ static CompressionCodec getCodec(String codecName) {
   public Status process() throws EventDeliveryException {
     Channel channel = getChannel();
     Transaction transaction = channel.getTransaction();
-    List<BucketWriter> writers = Lists.newArrayList();
     transaction.begin();
     try {
+      Set<BucketWriter> writers = new LinkedHashSet<>();
```
mpercy, line 359: nit: I don't think we need ordering, how about just HashSet<>()?

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1545,4 +1553,62 @@ private void testRetryRename(boolean closeSucceed)
...
+    final Set<BucketWriter> bucketWriters = new HashSet<>();
+    sink = new HDFSEventSink() {
+      @Override
+      BucketWriter initializeBucketWriter(String realPath, String realName, String lookupPath,
+                                          HDFSWriter hdfsWriter, WriterCallback closeCallback) {
+        BucketWriter bw = Mockito.spy(super.initializeBucketWriter(realPath, realName, lookupPath,
+            hdfsWriter, closeCallback));
+        try {
+          // create mock BucketWriters where the first append() succeeds but the
+          // the second call throws a BucketClosedException
```
mpercy, line 1577: I don't see any logic here to make this throw only on the second call. It seems like it would throw on every call. Maybe I'm missing something.

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
```diff
@@ -1545,4 +1553,62 @@ private void testRetryRename(boolean closeSucceed)
...
+      }
+    }).when(channel).take();
+
+    sink.setChannel(channel);
+    sink.start();
+
+    sink.process();
+
+    // It is expected that flush() method was called exactly once for every BucketWriter
+    for (BucketWriter bw : bucketWriters) {
```
mpercy: I think it would be good to add some more assertions: how many events were taken from the channel, how many events were successfully written to HDFS, and how many bucketWriters there are.

mpercy (changes requested):
Great find. I just have a few mostly minor comments.

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
adenes: Chaining the `do*` calls in Mockito's Stubber class ensures that the first call will call the real method and the second call throws the exception. 
For more details see the Stubber's javadoc: https://static.javadoc.io/org.mockito/mockito-core/1.9.0/org/mockito/stubbing/Stubber.html

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
adenes: I'm not sure but I vaguely remember some issue regarding to the ordering (cc. @simonati) so I would keep the `LinkedHashSet`. On the other hand not tests break with `HashSet`.

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
adenes: sure, will update the code

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSink.java
adenes: ok, will do.

adenes:
thanks @mpercy for the review, I'll update the PR soon.

adenes:
Hi @mpercy, I have updated the pull request based on your review. Could you please have a look on it again? Thanks.

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/HDFSEventSink.java
mpercy: OK

mpercy (approved):
+1 LGTM

PR #130, opened 2017-05-10
bugfix: when the file is rename, it's pos becomes to 0,and it will be send to…

leocook:
when the file is rename, it's pos change to 0,and it will be send to channel repeatedly.It looks like a bug to me,when I collect log what create by log4j etc.


chris-fast:
There is two bugs：
1.when rename the file,the file will be post to channel form pos 0 repeatedly;
2.If bug NO.1 is fixed，when rename the file,it will not post repeatedly, bug when restart flume agent, the file witch is  renamed will post to channel repeatedly.

chris-fast:
Is anyone here?

chris-fast:
hello ？

asfgit:

PR #131, opened 2017-05-15
FLUME-3092. Extend the FileChannel's monitoring metrics

adenes:
This patch adds the following new metrics to the FileChannel's counters:
- eventPutErrorCount: incremented if an IOException occurs during put operation.
- eventTakeErrorCount: incremented if an IOException or CorruptEventException occurs
  during take operation.
- checkpointWriteErrorCount: incremented if an exception occurs during checkpoint write.
- unhealthy: this flag represents whether the channel has started successfully
  (i.e. the replay ran without any problem), so the channel is capable for normal operation
- closed flag: the numeric representation (1: closed, 0: open) of the negated open flag.

### DIFF
--- a/flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
+++ b/flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
@@ -39,7 +39,6 @@ import org.apache.flume.channel.file.Log.Builder;
 import org.apache.flume.channel.file.encryption.EncryptionConfiguration;
 import org.apache.flume.channel.file.encryption.KeyProvider;
 import org.apache.flume.channel.file.encryption.KeyProviderFactory;
-import org.apache.flume.instrumentation.ChannelCounter;
 import org.apache.flume.channel.file.instrumentation.FileChannelCounter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -270,6 +269,7 @@ public class FileChannel extends BasicChannelSemantics {
     if (channelCounter == null) {
       channelCounter = new FileChannelCounter(getName());
     }
+    channelCounter.setUnhealthy(false);
   }
 
   @Override
@@ -277,25 +277,7 @@ public class FileChannel extends BasicChannelSemantics {
     LOG.info("Starting {}...", this);
     channelCounter.start();
     try {
-      Builder builder = new Log.Builder();
-      builder.setCheckpointInterval(checkpointInterval);
-      builder.setMaxFileSize(maxFileSize);
-      builder.setMinimumRequiredSpace(minimumRequiredSpace);
-      builder.setQueueSize(capacity);
-      builder.setCheckpointDir(checkpointDir);
-      builder.setLogDirs(dataDirs);
-      builder.setChannelName(getName());
-      builder.setUseLogReplayV1(useLogReplayV1);
-      builder.setUseFastReplay(useFastReplay);
-      builder.setEncryptionKeyProvider(encryptionKeyProvider);
-      builder.setEncryptionKeyAlias(encryptionActiveKey);
-      builder.setEncryptionCipherProvider(encryptionCipherProvider);
-      builder.setUseDualCheckpoints(useDualCheckpoints);
-      builder.setCompressBackupCheckpoint(compressBackupCheckpoint);
-      builder.setBackupCheckpointDir(backupCheckpointDir);
-      builder.setFsyncPerTransaction(fsyncPerTransaction);
-      builder.setFsyncInterval(fsyncInterval);
-      builder.setCheckpointOnClose(checkpointOnClose);
+      Builder builder = createLogBuilder();
       log = builder.build();
       log.replay();
       setOpen(true);
@@ -307,6 +289,7 @@ public class FileChannel extends BasicChannelSemantics {
           + channelNameDescriptor);
     } catch (Throwable t) {
       setOpen(false);
+      channelCounter.setUnhealthy(true);
       startupError = t;
       LOG.error("Failed to start the file channel " + channelNameDescriptor, t);
       if (t instanceof Error) {
@@ -318,6 +301,31 @@ public class FileChannel extends BasicChannelSemantics {
       channelCounter.setChannelCapacity(capacity);
     }
     super.start();
+  }
+
+  @VisibleForTesting
+  Builder createLogBuilder() {
+    Builder builder = new Log.Builder();
+    builder.setCheckpointInterval(checkpointInterval);
+    builder.setMaxFileSize(maxFileSize);
+    builder.setMinimumRequiredSpace(minimumRequiredSpace);
+    builder.setQueueSize(capacity);
+    builder.setCheckpointDir(checkpointDir);
+    builder.setLogDirs(dataDirs);
+    builder.setChannelName(getName());
+    builder.setUseLogReplayV1(useLogReplayV1);
+    builder.setUseFastReplay(useFastReplay);
+    builder.setEncryptionKeyProvider(encryptionKeyProvider);
+    builder.setEncryptionKeyAlias(encryptionActiveKey);
+    builder.setEncryptionCipherProvider(encryptionCipherProvider);
+    builder.setUseDualCheckpoints(useDualCheckpoints);
+    builder.setCompressBackupCheckpoint(compressBackupCheckpoint);
+    builder.setBackupCheckpointDir(backupCheckpointDir);
+    builder.setFsyncPerTransaction(fsyncPerTransaction);
+    builder.setFsyncInterval(fsyncInterval);
+    builder.setCheckpointOnClose(checkpointOnClose);
+    builder.setChannelCounter(channelCounter);
+    return builder;
   }
 
   @Override
@@ -449,12 +457,12 @@ public class FileChannel extends BasicChannelSemantics {
     private final FlumeEventQueue queue;
     private final Semaphore queueRemaining;
     private final String channelNameDescriptor;
-    private final ChannelCounter channelCounter;
+    private final FileChannelCounter channelCounter;
     private final boolean fsyncPerTransaction;
 
     public FileBackedTransaction(Log log, long transactionID,
                                  int transCapacity, int keepAlive, Semaphore queueRemaining,
-                                 String name, boolean fsyncPerTransaction, ChannelCounter
+                                 String name, boolean fsyncPerTransaction, FileChannelCounter
                                      counter) {
       this.log = log;
       queue = log.getFlumeEventQueue();
@@ -503,6 +511,7 @@ public class FileChannel extends BasicChannelSemantics {
         queue.addWithoutCommit(ptr, transactionID);
         success = true;
       } catch (IOException e) {
+        channelCounter.incrementEventPutErrorCount();
         throw new ChannelException("Put failed due to IO error "
             + channelNameDescriptor, e);
       } finally {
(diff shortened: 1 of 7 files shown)


flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/instrumentation/FileChannelCounter.java
```diff
@@ -36,4 +42,45 @@ public boolean isOpen() {
...
   }
+
+  @Override
+  public int getClosed() {
+    return open ? 0 : 1;
+  }
+
+  @Override
+  public int getUnhealthy() {
+    return unhealthy ? 1 : 0;
```
mpercy: It's sort of weird that the getter is an int but the setter is a boolean. Can we just make everything an int, including the underlying variable? Also, did you consider making this an error count instead of just 1/0 since that would provide additional information?

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelBase.java
```diff
@@ -41,6 +42,14 @@
   protected File uncompressedBackupCheckpoint;
   protected File compressedBackupCheckpoint;
 
+  public TestFileChannelBase() {
+    this(3);
```
mpercy: nit: magic number. Please add comment for why we want 3

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
```diff
@@ -0,0 +1,222 @@
...
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertTrue;
+
+public class TestFileChannelErrorMetrics extends TestFileChannelBase {
+
+  public TestFileChannelErrorMetrics() {
+    super(1);
```
mpercy: nit: magic number. Please add a comment why we want 1

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
```diff
@@ -0,0 +1,222 @@
...
+
+    long beforeCheckpointWrite = System.currentTimeMillis();
+
+    // first checkpoint should be written successfully -> the counter should remain 0
+    Thread.sleep(checkpointInterval + 100);
+    assertEquals(0, channel.getChannelCounter().getCheckpointWriteErrorCount());
+    assertTrue(new File(checkpointDir, "checkpoint").lastModified() > beforeCheckpointWrite);
+
+    FileUtils.deleteDirectory(baseDir);
+    Thread.sleep(checkpointInterval + 100);
```
mpercy: This is likely going to be flaky under system load. You could probably make it fail by running this test in a loop while also running stress -c 16 in another terminal.

I would suggest doing something like this:

```
FileUtils.deleteDirectory(baseDir);
final long TIMEOUT_MILLIS = 15000;
while (System.currentTimeMillis() < beforeCheckPointWrite + TIMEOUT_MILLIS) {
  if (channel.getChannelCounter().getCheckpointWriteErrorCount() > 0) {
    break;
  }
  Thread.sleep(10);
}
// the channel's directory has been deleted so the checkpoint write should have failed
assertThat("checkpointWriteErrorCount",
    channel.getChannelCounter().getCheckpointWriteErrorCount(),
    greaterThan(0));
```
You could probably even wrap that loop in a helper function and have it take a lambda expression, i.e.

```
void assertEventuallyTrue(String description, BooleanSupplier expr, long timeoutMillis) {
  long start = System.currentTimeMillis();
  while (System.currentTimeMillis() < start + timeoutMillis) {
    if (expr.getAsBoolean()) break;
    Thread.sleep(10); // Sleep for 10ms before each retry
  }
  assertTrue(desc, expr.getAsBoolean());
}

FileUtils.deleteDirectory(baseDir);
long timeoutMillis = 15000;
assertEventuallyTrue("checkpointWriteErrorCount > 0",
                     () -> channel.getChannelCounter().getCheckpointWriteErrorCount() > 0,
                     timeoutMillis);
```

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
```diff
@@ -0,0 +1,222 @@
...
+    tx.begin();
+    channel.take();
+    tx.commit();
+    tx.close();
+    assertEquals(1, channelCounter.getEventTakeAttemptCount());
+    assertEquals(1, channelCounter.getEventTakeSuccessCount());
+    assertEquals(0, channelCounter.getEventTakeErrorCount());
+
+    FileUtils.deleteDirectory(baseDir);
+    Thread.sleep(2 * usableSpaceRefreshInterval);
```
mpercy, line 89: This sleep will make this test potentially flaky. Can you make it more reliable? Perhaps using the assertEventuallyTrue() function I suggest in another comment below. Perhaps you could put the entire following block in a lambda, i.e.

```
  assertEventuallyTrue("putErrorCount > 0",
      () -> {
        Transaction tx = channel.getTransaction();
        tx.begin();
        try {
          channel.put(EventBuilder.withBody("test".getBytes()));
          tx.commit();
          tx.close();
        } catch (ChannelException ex) {
          log.info("ChannelException", ex);
        }
        return channelCounter.getEventPutErrorCount() > 0;
      },
      timeoutMillis);
```

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
```diff
@@ -0,0 +1,222 @@
...
+    } catch (ChannelException ex) {
+      putException = ex;
+    }
+    assertNotNull(putException);
+    assertTrue(putException.getCause() instanceof IOException);
+    assertEquals(3, channelCounter.getEventPutAttemptCount());
+    assertEquals(2, channelCounter.getEventPutSuccessCount());
+    assertEquals(1, channelCounter.getEventPutErrorCount());
+
+    ChannelException takeException = null;
```
mpercy, line 105: Please add comment like:

// This is guaranteed to throw an error if the above put() threw an error.

flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/instrumentation/FileChannelCounterMBean.java
```diff
@@ -23,4 +23,14 @@
 public interface FileChannelCounterMBean extends ChannelCounterMBean {
 
   boolean isOpen();
+
+  int getClosed();
+
+  int getUnhealthy();
+
+  long getEventPutErrorCount();
```
mpercy: Please add Javadoc for each metric in this interface. It should describe the purpose and behavior of the metric.

For example:

```
/** A count of the number of errors encountered while trying to put() onto the channel,
  * including IOExceptions and corruption-related errors.
  */
long getEventPutErrorCount();
```

mpercy (commented):
Looks good overall, just needs a few tweaks. Please see my review comments.

flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
```diff
@@ -503,6 +511,7 @@ protected void doPut(Event event) throws InterruptedException {
         queue.addWithoutCommit(ptr, transactionID);
         success = true;
       } catch (IOException e) {
+        channelCounter.incrementEventPutErrorCount();
```
mpercy, line 514: Don't we want to mark the channel unhealthy here too? And elsewhere that we increment event take / put error count as well as checkpoint write failure?

adenes:
Thanks for the review @mpercy, I've updated your pull request based on your comments.

flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/FileChannel.java
adenes: No, I explained it in more details in the comment for the `unhealthy` flag.

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
adenes: It won't be flaky, because the reason of the sleep is that the usable space is cached for the configured interval (see: https://github.com/apache/flume/blob/trunk/flume-ng-channels/flume-file-channel/src/main/java/org/apache/flume/channel/file/LogFile.java#L162-L162), so the test has simply wait for it to get invalidated.
There is no other thread or any other condition the test has to wait for and which might not happen during the sleep.

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
```diff
@@ -159,8 +161,8 @@ public void testCorruptEventTaken() throws Exception {
 
   @Test
   public void testCheckpointWriteErrorCount() throws Exception {
-    int checkpointInterval = 1000;
-    FileChannel channel = createFileChannel(Collections.singletonMap(
+    int checkpointInterval = 1500;
```
adenes: I increased this to 1500 because 1000 made the test flaky: it might have happened that the checkpoint was created in the same second when the checkpoint writer thread updated it which caused the `lastModified` check to fail on OSes/file systems where the lastmod timestamp has second granularity.

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
adenes: Thank you for your suggestion. I created a custom class as we are on Java 7, but the logic itself is the same.

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
simonati: I agree with @adenes on this. This cache is not refreshed by a background thread but essentially by the channel.put() call from the test itself.

flume-ng-channels/flume-file-channel/src/test/java/org/apache/flume/channel/file/TestFileChannelErrorMetrics.java
```diff
@@ -0,0 +1,247 @@
...
+    assertEquals(1, channel.getChannelCounter().getUnhealthy());
+    assertEquals(1, channel.getChannelCounter().getClosed());
+    assertFalse(channel.getChannelCounter().isOpen());
+  }
+
+  private interface BooleanPredicate {
+    boolean get();
+  }
+
+  private static void assertEventuallyTrue(String description, BooleanPredicate expression,
```
simonati, line 237: Nice one!

simonati (approved):
lgtm. unit tests are passing for me (5 consecutive iterations).

mvn test -pl flume-ng-channels/flume-file-channel

Results :
Tests run: 206, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
[INFO] Total time: 05:09 min
[INFO] Finished at: 2017-05-16T16:43:04+02:00

mpercy (approved):
+1, LGTM

Thanks for the follow-up Denes. I am about to commit this.

mpercy (approved):

PR #133, opened 2017-05-22
[FLUME-3096] add Google Cloud Pub/Sub Source

stakafum:


asfgit:

martinKindall:
What happened to this PR? Is there another way to use Flume with Pub/Sub?

rgoers:
If this is created it should be to its own Flume repo. The build already takes too long and is hard to maintain due to all the third party dependencies.

PR #136, opened 2017-05-24
[FLUME-3098] Add Amazon Kinesis Stream Sink

stakafum:


asfgit:

rgoers:
If this is created it should be to its own Flume repo. The build already takes too long and is hard to maintain due to all the third party dependencies.

PR #138, opened 2017-05-31
[FLUME-3090]Cassandra Sink module.

lukangping:
Here is the basic module for cassandra sink, which is used in our product code for collecting specific log event into the cassandra table.
There are 2 event serialization options: JsonCassandraEventSerializer and RegexCassandraEventSerializer.  For the property configuration, should be specified as:

a1.sinks.k1.type = cassandra
a1.sinks.k1.serializer = org.apache.flume.sink.cassandra.JsonCassandraEventSerializer
a1.sinks.k1.cassandra.contactPoints = 172.30.10.241
a1.sinks.k1.cassandra.username = test
a1.sinks.k1.cassandra.password = test
a1.sinks.k1.cassandra.keyspace =test_keyspace
a1.sinks.k1.cassandra.table = test_table
a1.sinks.k1.datetime.format = yyyy/MM/dd HH:mm:ss

Cheers.


asfgit:

rgoers:
If this is added, it should be to a separate repo. We are struggling to maintain all the various third party integrations that are already in Flume, much less add new ones, and the build already is too long. In addition, I suspect that the version of Cassandra being used here is probably no longer current.

rgoers:
> If this is added, it should be to a separate repo. We are struggling to maintain all the various third party integrations that are already in Flume, much less add new ones, and the build already is too long.

PR #139, opened 2017-06-20
HttpSink. Fix bad response handling. 

filippovmn:
After a bad response, connection.getInputStream () returns null. I'm adding a check for this. If the answer is bad - try to close ErrorStream.


bessbd:
@filippovmn : thank you for the pull request!

Is there an easy way to reproduce the faulty behavior in order to validate the patch?
Do you think there is an easy way to add a unit test for this?

filippovmn:
@bessbd
OK. I think it's not difficult

filippovmn:
@bessbd
I added a test that checks the ability to ignore some types of errors (in this case: 401) using the configuration.

bessbd:
I've run the tests and it looks fine to me.
I'll leave some time for others to review and leave comments if they have any.

bessbd:
LGTM, +1
I'm about to commit this

bessbd:
@filippovmn : thank you for the patch!

PR #140, opened 2017-06-26
small fix correct typo

TeddyBear1314:


simonati:
TravisCI failed since it incorrectly used jdk1.7.0_80 for the build. #142 fixed that issue. 

Could you please update this pull request to retrigger Travis? (`git commit --amend` then a `git push --force` would be sufficient).  

You may add some context to the commit message eg: 
```
Fix a typo in Sink metrics. 

API breaking change since the name of a counter changed.
```

mcsanady (approved):
Please make this change at flume-ng-doc/sphinx/FlumeUserGuide.rst:4537

mcsanady:
This also appears in FlumeUserGuide.rst 
I wonder if it is autogenerated, and if it is, how to regenerate.

TeddyBear1314:
Um...., I just started reading the Flume source code, really not familiar with the question you raised @mcsanady

mcsanady:
OK, I see, our *.rst files are not generated from the source. Your typo change should be applied too.
So please make this change at flume-ng-doc/sphinx/FlumeUserGuide.rst:4537

mcsanady:
Since this change is a "breaking change", please open a jira ticket too, to make it easier to track this change to get it appear in the next release notes.

simonati:
Agree with @mcsanady on this, we need to track that it might cause problems for exiting users. I assume release notes would be the right place. For that I think the following would be sufficient (afaik flume generates release notes from jira):

0. access to flume jira (https://issues.apache.org/jira/projects/FLUME)
1. create a new issue
2. put a label on it indicating that it is a breaking change and 
3. fill out its release notes informing existing adopters what to look for during upgrade

Please let us know if you need any help with it. Generally you need to became a contributor to access the jira which requires subscribing to the dev@ list plus a mail request to the community stating something like "I would like to became a contributor, please grant me access to the jira". It is completely fair to say no at this time if you don't want the associated hassle for this single change but I hope that many more would come so you would have to do eventually.

asfgit:

PR #141, opened 2017-06-29
FLUME-2752 Fix AvroSource startup resource leaks

simonati:
Cleanup after Netty initialisation failes (call this.stop())
Make sure this.stop() releases the resources and end up the component in
a LifecycleAware.STOPPED state
Added junit tests to cover the invalid host and used port scenarios

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
@@ -20,7 +20,6 @@
 package org.apache.flume.source;
 
 import com.google.common.base.Preconditions;
-import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.avro.ipc.NettyServer;
 import org.apache.avro.ipc.NettyTransceiver;
@@ -156,6 +155,7 @@ public class AvroSource extends AbstractSource implements EventDrivenSource,
   private boolean enableIpFilter;
   private String patternRuleConfigDefinition;
 
+  private NioServerSocketChannelFactory socketChannelFactory;
   private Server server;
   private SourceCounter sourceCounter;
 
@@ -233,14 +233,20 @@ public class AvroSource extends AbstractSource implements EventDrivenSource,
   public void start() {
     logger.info("Starting {}...", this);
 
-    Responder responder = new SpecificResponder(AvroSourceProtocol.class, this);
+    try {
+      Responder responder = new SpecificResponder(AvroSourceProtocol.class, this);
 
-    NioServerSocketChannelFactory socketChannelFactory = initSocketChannelFactory();
+      socketChannelFactory = initSocketChannelFactory();
 
-    ChannelPipelineFactory pipelineFactory = initChannelPipelineFactory();
+      ChannelPipelineFactory pipelineFactory = initChannelPipelineFactory();
 
-    server = new NettyServer(responder, new InetSocketAddress(bindAddress, port),
-          socketChannelFactory, pipelineFactory, null);
+      server = new NettyServer(responder, new InetSocketAddress(bindAddress, port),
+              socketChannelFactory, pipelineFactory, null);
+    } catch (org.jboss.netty.channel.ChannelException nce) {
+      logger.error("Avro source {} startup failed. Cannot initialize Netty server", getName(), nce);
+      stop();
+      throw new FlumeException(nce);
+    }
 
     connectionCountUpdater = Executors.newSingleThreadScheduledExecutor();
     server.start();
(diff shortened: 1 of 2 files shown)


simonati:
Tests executed (and passed):
mvn test -DfailIfNoTests=false -fae -Dtest=TestAvroSource,TestAvroSink,TestDefaultSourceFactory,TestLoadBalancingLog4jAppender,TestLog4jAppender,TestLog4jAppenderWithAvro

(all tests where AvroSource was referenced)

simonati:
TravisCI failed since it incorrectly used jdk1.7.0_80 for the build. Java target version was bumped to 1.8 some time ago. Please merge #142 first for fixing that.

adenes:
Thank you for the patch @simonati. I have committed the #142 pull request, could you please push a new commit for this PR to retrigger the Travis CI build? Thanks.

flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
```diff
@@ -300,28 +306,30 @@ public ChannelPipeline getPipeline() throws Exception {
...
     logger.info("Avro source {} stopping: {}", getName(), this);
 
-    server.close();
+    if (server != null) {
+      server.close();
+      try {
+        server.join();
+        server = null;
+      } catch (InterruptedException e) {
+        logger.info("Avro source " + getName() + ": Interrupted while waiting " +
```
adenes, line 315: I think the`Thread.currentThread().interrupt();` should be called here to set the interrupt flag of the thread.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
```diff
@@ -97,20 +96,19 @@ public void testLifecycle() throws InterruptedException {
...
-
-        context.put("port", String.valueOf(selectedPort = 41414 + i));
-        context.put("bind", "0.0.0.0");
 
-        Configurables.configure(source, context);
+      Context context = new Context();
 
+      context.put("port", String.valueOf(selectedPort = 41414 + i));
+      context.put("bind", "0.0.0.0");
+      // Invalid configuration may throw a FlumeException which has to be expected in the callers
```
adenes, line 104: Could you please explain what this comment means? Or is it related to the changes you made, i.e. removing the configuration part from the `try-catch` block?

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
```diff
@@ -129,6 +127,64 @@ public void testLifecycle() throws InterruptedException {
...
+      context.put("port", String.valueOf(port));
+      context.put("bind", loopbackIPv4);
+      // Invalid configuration may throw a FlumeException which has to be expected in the callers
+      Configurables.configure(source, context);
+      try {
+        source.start();
+        Assert.fail("Expected an exception during startup caused by binding on a used port");
+      } catch (FlumeException e) {
+        logger.info("Received an expected exception.", e);
+        Assert.assertTrue("Expected a bind related root cause",
```
adenes: Wouldn't it be useful to do an `instanceof` check for `e.getCause()`?

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
```diff
@@ -129,6 +127,64 @@ public void testLifecycle() throws InterruptedException {
...
+    context.put("bind", invalidHost);
+    // Invalid configuration may throw a FlumeException which has to be expected in the callers
+    Configurables.configure(source, context);
+
+    try {
+      source.start();
+      Assert.fail("Expected an exception during startup caused by binding on a invalid host");
+    } catch (FlumeException e) {
+      logger.info("Received an expected exception.", e);
+      Assert.assertTrue("Expected a bind related root cause",
```
adenes: Same as above (line 150)

adenes (changes requested):
Thank you for the patch @simonati , I had some minor comments, otherwise looks good to me.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
simonati: I think that should be considered as internal implementation detail (the situation that we use Netty so it throws a netty related exception should not be considered by any user of the AvroSource). I think the "API" here is a FlumeException with the relevant message.

flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
simonati: Sure thing.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
simonati: Yes it is related to that change. Can you please recommend a better wording?

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
adenes: I kind of agree, but with the current implementation (throwing `new FlumeException(nce)`) the message itself is an implementation detail as well, because it comes directly from the netty code.
I'd suggest to explicitly define the message when throwing the `FlumeException`. Wdyt?

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
simonati: I think the same answer applies.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
adenes: I'm not sure it is needed here at all, without knowing the previous version of the code it cannot be understood. I'd simply remove it.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
simonati: Please let me know if you think otherwise.

simonati: > I'd suggest to explicitly define the message

Good point.

simonati: Does this sound good?

throw new FlumeException("Failed to set up server socket", nce);

adenes: yes, thx

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
simonati: With one exception I removed this comment. In that occurrence the helper function is called by different tests and some of the tests expect a FlumeException to be thrown caused by invalid configuration. It could have saved me some time during implementation that `configure()` can also raise a FlumeException so I would like to make life of future devs a bit easier. Also I rephrased a little the comment.

adenes:
Thanks for updating the patch, LGTM 👍

PR #142, opened 2017-06-29
Compile with jdk8 on travisCI

simonati:


adenes:
Thanks @simonati for the patch, LGTM +1

PR #143, opened 2017-06-30
Facelift AvroSource and test using lambdas

simonati:
Make avro ip filter tests more reliable by checking whether the caught exception is really what the test expected
use lambda instead of anonymous classes to make the code shorter

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
@@ -253,14 +253,9 @@ public class AvroSource extends AbstractSource implements EventDrivenSource,
     sourceCounter.start();
     super.start();
     final NettyServer srv = (NettyServer)server;
-    connectionCountUpdater.scheduleWithFixedDelay(new Runnable() {
-
-      @Override
-      public void run() {
-        sourceCounter.setOpenConnectionCount(
-                Long.valueOf(srv.getNumActiveConnections()));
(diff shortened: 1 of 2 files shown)


simonati:
Testing done: mvn test -DfailIfNoTests=false -fae -Dtest=TestAvroSource,TestAvroSink,TestDefaultSourceFactory,TestLoadBalancingLog4jAppender,TestLog4jAppender,TestLog4jAppenderWithAvro

flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
```diff
@@ -292,12 +287,7 @@ private ChannelPipelineFactory initChannelPipelineFactory() {
         keystorePassword, keystoreType, enableIpFilter,
         patternRuleConfigDefinition);
     } else {
-      pipelineFactory = new ChannelPipelineFactory() {
-        @Override
-        public ChannelPipeline getPipeline() throws Exception {
-          return Channels.pipeline();
-        }
-      };
+      pipelineFactory = () -> Channels.pipeline();
```
adenes: This can be replaced by method reference, i.e. `pipelineFactory = Channels::pipeline`

adenes (changes requested):
Thanks for the patch, I had one comment, otherwise it looks good to me.

adenes (approved):
Thanks for the update, LGTM, 👍

PR #144, opened 2017-07-08
Flume-2786. Fixed nested monitor lockout with polling trial of application lock.

gangken:
This issue was discussed at [FLUME-2786](https://issues.apache.org/jira/browse/FLUME-2786), and I'd uploaded a patch for 1.6.0. Now I tried to apply the patch to trunk.

### DIFF
--- a/flume-ng-node/src/main/java/org/apache/flume/node/Application.java
+++ b/flume-ng-node/src/main/java/org/apache/flume/node/Application.java
@@ -52,6 +52,10 @@ import java.util.Locale;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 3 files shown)


flume-ng-node/src/main/java/org/apache/flume/node/Application.java
```diff
@@ -73,25 +79,53 @@ public Application() {
...
+          try {
+            stopAllComponents();
+            startAllComponents(conf);
+          } finally {
+            lock.unlock();
+          }
+          return;
+        }
+      } catch (InterruptedException e) {
+        // Ignore
```
asdf2014: Maybe this `catch` clause should change into the follow code style...
```java
catch (InterruptedException ignored)
```

gangken: Hi, @asdf2014 Thanks for your suggestion.

Sorry that I'v missed to reset the interrupted flag, so I log a warning message and interrupt the current thread to handle this exception.

asdf2014 (approved):

asdf2014:
LGTM.

PR #146, opened 2017-07-19
FLUME-3128: Fix TestHDFSEventSinkOnMiniCluster.java due to incompatib…

mcsanady:
…le changes in hadoop-hdfs lib.

Our test in flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestHDFSEventSinkOnMiniCluster.java
Uses the getLeaseByPath(String src) method of Lease class which is to be replaced by getLeaseForPath(Namenode n, String src) method.

This change is to make Flume compatible after the hadoop change is made effective.

See https://issues.apache.org/jira/browse/HDFS-6757


mcsanady:
It won't work. Not compatible with hadoop's currently used version.

PR #147, opened 2017-07-20
[FLUME-3133] Add a 'ipHeader' config in both tcp and udp syslog source

lingjinjiang:
When I use the syslog source, I use the "host" header to tell where the event come from.However, when change the format of syslog service, the "host" header willl be missing.

So I add a new config named "ipHeader" which looks like the "portHeader" in Multiport Syslog TCP Source.
When this config is specified, a header will be added.

For more detail: 
https://issues.apache.org/jira/browse/FLUME-3133

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/MultiportSyslogTCPSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/MultiportSyslogTCPSource.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.util.List;
@@ -66,6 +67,7 @@ public class MultiportSyslogTCPSource extends AbstractSource implements
   private int batchSize;
   private int readBufferSize;
   private String portHeader;
+  private String ipHeader;
   private SourceCounter sourceCounter = null;
   private Charset defaultCharset;
   private ThreadSafeDecoder defaultDecoder;
@@ -137,6 +139,9 @@ public class MultiportSyslogTCPSource extends AbstractSource implements
     portHeader = context.getString(
             SyslogSourceConfigurationConstants.CONFIG_PORT_HEADER);
 
+    ipHeader = context.getString(
+            SyslogSourceConfigurationConstants.CONFIG_IP_HEADER);
+
     readBufferSize = context.getInteger(
         SyslogSourceConfigurationConstants.CONFIG_READBUF_SIZE,
         SyslogSourceConfigurationConstants.DEFAULT_READBUF_SIZE);
@@ -166,8 +171,8 @@ public class MultiportSyslogTCPSource extends AbstractSource implements
     acceptor.getSessionConfig().setIdleTime(IdleStatus.BOTH_IDLE, 10);
 
     acceptor.setHandler(new MultiportSyslogHandler(maxEventSize, batchSize,
-        getChannelProcessor(), sourceCounter, portHeader, defaultDecoder,
-        portCharsets, keepFields));
+        getChannelProcessor(), sourceCounter, portHeader, ipHeader,
+        defaultDecoder, portCharsets, keepFields));
 
     for (int port : ports) {
       InetSocketAddress addr;
@@ -217,6 +222,7 @@ public class MultiportSyslogTCPSource extends AbstractSource implements
     private final int batchSize;
     private final SourceCounter sourceCounter;
     private final String portHeader;
+    private final String ipHeader;
     private final SyslogParser syslogParser;
     private final LineSplitter lineSplitter;
     private final ThreadSafeDecoder defaultDecoder;
@@ -225,7 +231,7 @@ public class MultiportSyslogTCPSource extends AbstractSource implements
 
     public MultiportSyslogHandler(int maxEventSize, int batchSize,
         ChannelProcessor cp, SourceCounter ctr, String portHeader,
-        ThreadSafeDecoder defaultDecoder,
+        String ipHeader, ThreadSafeDecoder defaultDecoder,
         ConcurrentMap<Integer, ThreadSafeDecoder> portCharsets,
         Set<String> keepFields) {
       channelProcessor = cp;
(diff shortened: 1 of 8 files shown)


bessbd:
Hi @lingjinjiang ,

Thank you for the pull request!

If I'm not mistaken, this change adds new configuration options.
Can you please add the necessary modifications to [FlumeUserGuide.rst](https://github.com/apache/flume/blob/trunk/flume-ng-doc/sphinx/FlumeUserGuide.rst)?


Thank you,

Donat

lingjinjiang:
Hi @bessbd ,
Thanks for your review.
I have updated the flume user guide.

flume-ng-core/src/main/java/org/apache/flume/source/MultiportSyslogTCPSource.java
```diff
@@ -137,6 +139,9 @@ public void configure(Context context) {
     portHeader = context.getString(
             SyslogSourceConfigurationConstants.CONFIG_PORT_HEADER);
 
+    ipHeader = context.getString(
+            SyslogSourceConfigurationConstants.CONFIG_IP_HEADER);
```
turcsanyip, line 143: The existing port header contains a _local_ port, while the new port header is the _remote_ IP, so they belong to two different network interfaces.
To make it clear, I would rename the new header to 'sourceIpHeader' or 'senderIpHeader'.

(The local IP could also be stored in a header for the case when the Flume host contains multiple network cards, and this IP would belong to the existing port header and could be called ipHeader simply.)

flume-ng-core/src/main/java/org/apache/flume/source/MultiportSyslogTCPSource.java
```diff
@@ -287,6 +294,16 @@ public void messageReceived(IoSession session, Object message) {
         decoder = portCharsets.get(port).get();
       }
 
+      //get ip from session
+      String ip = "";
+      SocketAddress socket = session.getRemoteAddress();
```
turcsanyip, line 299: It is an _address_ not a _socket_, so _(***)address_ name would be better for this variable.

flume-ng-core/src/main/java/org/apache/flume/source/SyslogSourceConfigurationConstants.java
```diff
@@ -74,6 +74,9 @@
   public static final String CONFIG_KEEP_FIELDS_TIMESTAMP = "timestamp";
   public static final String CONFIG_KEEP_FIELDS_HOSTNAME = "hostname";
 
+  public static final String CONFIG_IP_HEADER = "ipHeader";
+  public static final String DEFAULT_IP_HEADER = "ip";
```
szaboferee, line 78: This is only used in tests so it should be defined there and not in the production code

flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -1609,6 +1610,7 @@ keepFields            none              Setting this to 'all' will preserve the
                                         fields can be included: priority, version,
                                         timestamp, hostname. The values 'true' and 'false'
                                         have been deprecated in favor of 'all' and 'none'.
+ipHeader              --                If specified, the IP will be stored in the header of each event using the header name specified here. This allows for interceptors and channel selectors to customize routing logic based on the incoming host.
```
turcsanyip, line 1613: "the IP will be stored": please specify which IP
"incoming host": it might be better to use _source_ or _sender_ instead of incoming

flume-ng-core/src/test/java/org/apache/flume/source/TestMultiportSyslogTCPSource.java
```diff
@@ -383,4 +386,84 @@ public void testPortCharsetHandling() throws UnknownHostException, Exception {
...
+    context.put(SyslogSourceConfigurationConstants.CONFIG_IP_HEADER,
+            TEST_IP_HEADER);
+
+    source.configure(context);
+    source.start();
+
+    //create a socket to send a test event
+    Socket syslogSocket;
+    syslogSocket = new Socket(
+            InetAddress.getLocalHost(), getFreePort());
```
szaboferee, line 424: this breaks as the client tries to connect to a new random port and not to the port the server is listening on.

flume-ng-core/src/main/java/org/apache/flume/source/MultiportSyslogTCPSource.java
szaboferee, line 299: This could be checked and cast to an InetSocketAddress and then getAddress().getHostAddress() to retrieve the IP.
even better, based on a flag we could retrieve the hostname as IP addresses tend to be temporary nowadays.

szaboferee (changes requested):
Thanks for the contribution @lingjinjiang !
Nice pull request, I appreciate the tests and the documentation.
I have found some minor issues that need to be addressed before I could commit your change.
Let me know if you need any help.

flume-ng-core/src/test/java/org/apache/flume/source/TestMultiportSyslogTCPSource.java
turcsanyip, line 424: This does not work with the current trunk version.
Please rebase the PR onto trunk and fix the test failure.

turcsanyip (changes requested):
The general logic looks good to me.
There is a test failure with the current trunk version, please fix it.
I would suggest some changes / refinements regarding the header / variable names, and in the User Guide as well (please see my inline comments).

majorendre:
Hi @lingjinjiang,

Could you please help me to understand the problem this PR solves?
"However, when the syslog format is changed, the "host" header will miss."
Maybe the SyslogParser needs fixing.

Thanks,
Endre

asfgit:

flume-ng-core/src/main/java/org/apache/flume/source/MultiportSyslogTCPSource.java
majorendre: @szaboferee I agree
This is how the "port" is retrieved
290:      int port = ((InetSocketAddress) session.getLocalAddress()).getPort();
This seems much safer to me:
      String hostName = ((InetSocketAddress) session.getRemoteAddress()).getHostName();
      String ipAddress = ((InetSocketAddress) session.getRemoteAddress()).getAddress().getHostAddress();

turcsanyip:
Hi @lingjinjiang,
As there was no answer from your side, I took over the issue. Your change has been rebased onto the current trunk (as there have been a lot of changes in the meantime) and the review comments have been implemented.
For further details plase see https://github.com/apache/flume/pull/234
This PR can be closed.

PR #151, opened 2017-08-11
FLUME-3144. Improve Log4jAppender's performance by allowing logging collection of messages

adenes:
`Log4jAppender` treats `Collection` messages as a special case making it possible to log collection of events in one Log4j log call. The appender sends these events to the receiving Flume instance as one batch with the `rpcClient.appendBatch()` method.

New tests added:
- logging of a collection of multiple strings, expect one `appendBatch()` call.
- logging one string, expect one `append()` call.
- logging of a singleton collection of a string, expect one `append()` call.
- logging of an empty collection, expect that neither `append()` nor `appendBatch()` was called.
- logging of a `String` and an arbitrary object in one collection with `AvroReflectionEnabled` set to `true`
- logging of a `String` and an arbitrary object in one collection with `AvroReflectionEnabled` set to `false`

### DIFF
--- a/flume-ng-clients/flume-ng-log4jappender/pom.xml
+++ b/flume-ng-clients/flume-ng-log4jappender/pom.xml
@@ -87,6 +87,12 @@ limitations under the License.
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.mockito</groupId>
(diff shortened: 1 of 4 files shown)


szaboferee (approved):
LGTM :+1:

mcsanady (approved):
Ship it!

PR #152, opened 2017-08-15
FLUME-3132 Upgrade tomcat jasper library dependencies

szaboferee:
excluding vulnerable dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,6 +209,16 @@ limitations under the License.
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-hdfs</artifactId>
             <version>${hadoop.version}</version>
+            <exclusions>
+              <exclusion>
+                <groupId>tomcat</groupId>
+                <artifactId>jasper-compiler</artifactId>
+              </exclusion>
+              <exclusion>
+                <groupId>tomcat</groupId>
+                <artifactId>jasper-runtime</artifactId>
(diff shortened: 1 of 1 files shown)


mcsanady (approved):
Looks good to me.

simonati (commented):
With this change Jasper will be only a test scope dependency which is better than what we have now. 

Is jasper really needed to run our tests?

szaboferee:
In my understanding Jasper is not needed anywhere, however it is a transitive dependency from hadoop   and hadoop is a dependency in some test cases

simonati:
If it is not needed for our tests then do you think it is worth excluding it from the test scope as well?

szaboferee:
I removed it from the test scope as well.
the new dependency diff is:
```
71d70
< [INFO]    commons-el:commons-el:jar:1.0:compile
532,536d530
< [INFO]    tomcat:jasper-compiler:jar:5.5.23:runtime
< [INFO]    tomcat:jasper-compiler:jar:5.5.23:test
< [INFO]    tomcat:jasper-runtime:jar:5.5.23:compile
< [INFO]    tomcat:jasper-runtime:jar:5.5.23:runtime
< [INFO]    tomcat:jasper-runtime:jar:5.5.23:test
```

The build with tests was successful

simonati (approved):
LGTM

PR #153, opened 2017-08-15
FLUME-3131 Upgrade spring framework library dependencies

szaboferee:
they cannot be upgraded because they are  activemq dependencies. they were moved to test scope

### DIFF
--- a/flume-ng-sources/flume-jms-source/pom.xml
+++ b/flume-ng-sources/flume-jms-source/pom.xml
@@ -71,9 +71,14 @@ limitations under the License.
     </dependency>
 
     <dependency>
+      <groupId>javax.jms</groupId>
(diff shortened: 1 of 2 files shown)


flume-ng-sources/flume-jms-source/pom.xml
```diff
@@ -70,10 +70,15 @@ limitations under the License.
       <scope>test</scope>
     </dependency>
 
+  <dependency>
```
adenes: nit: could you please fix the indentation here?

adenes (changes requested):
One minor change requested, otherwise LGTM

simonati (commented):
Besides indentation pointed out by Denes it looks good to me.

PR #154, opened 2017-08-15
FLUME-3112 Upgrade jackson-core library dependency

szaboferee:
Upgrading in dependency management and removing unused ones

### DIFF
--- a/flume-ng-node/pom.xml
+++ b/flume-ng-node/pom.xml
@@ -151,16 +151,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.codehaus.jackson</groupId>
-      <artifactId>jackson-core-asl</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.codehaus.jackson</groupId>
-      <artifactId>jackson-mapper-asl</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-framework</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@ limitations under the License.
     <bundle-plugin.version>2.3.7</bundle-plugin.version>
     <checkstyle.plugin.version>2.17</checkstyle.plugin.version>
     <checkstyle.tool.version>6.19</checkstyle.tool.version>
+    <codehaus.jackson.version>1.9.3</codehaus.jackson.version>
     <commons-cli.version>1.2</commons-cli.version>
     <commons-codec.version>1.8</commons-codec.version>
     <commons-collections.version>3.2.2</commons-collections.version>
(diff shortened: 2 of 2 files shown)


adenes (approved):
LGTM, 👍

mcsanady (approved):
LGTM, Thank you Ferenc

simonati (commented):
Why do we rely on the other jackson version (1.9.3)? Can we either merge these usages or make sure the version property has a self explaining name?

szaboferee:
The simple jackson.version is for the 1.x from the org.codehaus.jackson group. I kept it because we have transitive dependencies for these artifacts. These usages cannot be merged( different groupId).

The jackson2.version is the 2.x from the com.fasterxml.jackson.core group. This is also for transitive dependencies. The 2 in jackson2 is the mayor version. For me it was self explaining. Do you have any suggestion for the name @simonati ?

szaboferee:
@simonati I've changed the version names and added another jackson dependency to the patch. Let me know if it looks ok for you

simonati:
Thanks @szaboferee! The diff looks good to me. Thanks for making the properties self explain and handling all of the jackson deps (instead of making the change only for jackson-core). I'm running tests of flume now to have a double verification. Will get back to you once it finished.

simonati:
{{mvn clean install -Dsurefire.rerunFailingTestsCount=5}} fails with 
```
[ERROR] Failed to execute goal on project flume-ng-morphline-solr-sink: Could not resolve dependencies for project org.apache.flume.flume-ng-sinks:flume-ng-morphline-solr-sink:jar:1.8.0-SNAPSHOT: Failed to collect dependencies at org.kitesdk:kite-morphlines-all:pom:1.0.0 -> org.kitesdk:kite-morphlines-json:jar:1.0.0 -> com.fasterxml.jackson.core:jackson-databind:jar:2.8.9 -> com.fasterxml.jackson.core:jackson-annotations:jar:${jackson2.version}: Failed to read artifact descriptor for com.fasterxml.jackson.core:jackson-annotations:jar:${jackson2.version}: Could not transfer artifact com.fasterxml.jackson.core:jackson-annotations:pom:${jackson2.version} from/to com.cloudera.releases (https://repository.cloudera.com/artifactory/cloudera-repos/): Failed to transfer file: https://repository.cloudera.com/artifactory/cloudera-repos/com/fasterxml/jackson/core/jackson-annotations/$%7Bjackson2.version%7D/jackson-annotations-$%7Bjackson2.version%7D.pom. Return code is: 400 , ReasonPhrase:Bad Request. -> [Help 1]
```

I think the problem is caused by the remaining reference to the {{jackson2.version}} property:
```
pom.xml
1119:        <groupId>com.fasterxml.jackson.core</groupId>
1120:        <artifactId>jackson-annotations</artifactId>
1121:        <version>${jackson2.version}</version>
1125:        <groupId>com.fasterxml.jackson.core</groupId>
1126:        <artifactId>jackson-core</artifactId>
1127:        <version>${jackson2.version}</version>
```

szaboferee:
Hi @simonati, thanks for testing it, my test runs were successful. It looks like that you had some mix of an earlier version of this pull request, because the version property name is fasterxml.jackson.version and not jackson2.version as in your error message. Also make sure, that you use the proper settings.xml, because the cloudera mirror does not necessarily contains every artifact that is needed for the flume build. 

try with `mvn clean install -Dsurefire.version=2.20 -Dsurefire.rerunFailingTestsCount=5`

simonati:
> looks like that you had some mix of an earlier version of this pull request,

Indeed that was the case. Thanks!

simonati (approved):
Tests (junit + flume-ng-tests) passed for me as well. Looks good to me!

PR #155, opened 2017-08-21
FLUME-3115 Update netty library

szaboferee:


pom.xml
```diff
@@ -100,7 +100,8 @@ limitations under the License.
     <mvn-surefire-plugin.version>2.14.1</mvn-surefire-plugin.version>
     <mvn-surefire-plugin-old.version>2.12.3</mvn-surefire-plugin-old.version>
     <mvn-surefire-plugin-old-morphline.version>2.12.4</mvn-surefire-plugin-old-morphline.version>
-    <netty.version>3.9.4.Final</netty.version>
+    <netty.version>3.10.6.Final</netty.version>
+    <netty-all.version>4.1.14.Final</netty-all.version>
```
mcsanady: Today the latest version is 4.1.17.final
Do we have a reason to use the older one?

mcsanady (changes requested):
Thank you for this contribution.
Since it was open some versions have released.
Consider using them.
Otherwise seems look to me.

pom.xml
szaboferee: Thanks for pointing it out. I will update it

mcsanady (approved):

PR #156, opened 2017-08-22
FLUME-3152 Add Flume Metric for Backup Checkpoint Errors

szaboferee:
Added a new metric to the file channel

new test case added
build with tests was successful


adenes (changes requested):
Hi @szaboferee, thank you for the patch. All in all it looks good to me, I have only one comment: could you please add a more "extended" test? Like the ones for the other error counters.
Thanks.

adenes:
@szaboferee , thanks for the added test case.
The change looks good to me, 👍

PR #158, opened 2017-08-24
FLUME-2698 Upgrade Jetty Version

tmgstevens:
This patch upgrades the HTTPSource and HTTP Metrics server to Jetty 9.4. It also adds the ability to further configure the Jetty server and also run in multi-threaded mode. Finally, registered Jetty with the MBeans server to allow fine-grained performance tuning/monitoring if required.


szaboferee:
I made some performance testing and the new implementation with the current default settings is much slower.

I will provide a detailed report and some help to reproduce it

szaboferee:
The slowness came from a too small default heap size, my measuring results are on the ticket.
Jetty 9 came out better at the end.

szaboferee (approved):

mcsanady:
Looks Good, Thank you for the patch

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+import org.slf4j.Logger;
+
+/**
+ * Utility class to enable runtime configuration of Java objects using provided
+ * Flume context objects (or equivalent). The methods use reflection to identify
+ * Fields on the configurable object and then looks for matching properties in
+ * the provided properties bundle.
+ *
+ */
+public class FlumeBeanConfigurator {
```
szaboferee, line 36: I'm a little bit concerned about this FlumeBeanConfigurator class, because if it is not used with caution it could expose setters we might would not like to expose. I see why it is easier to use a util like this with the jetty configuration. We should consider moving this somewhere so only the HttpSource  could see and use it.

Also some question that could be documented is :
When the user sets both `keystorePassword` and `SslContextFactory.KeyStorePassword` which one will be set?

szaboferee (changes requested):

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
tmgstevens: Clarified the ``keystore`` bit in the docs and actually changed the order of precedence.

Regarding the ``FlumeBeanConfigurator`` I'm afraid I don't agree that we could expose anything that wasn't already exposed. The Configurator only uses public setters, so cannot set any property that wasn't already available to be set - developers obviously need to use it with their eyes open.

I've added a set of Unit tests to test the functionality of this, including testing that private setters cannot be modified. Hopefully this satisfies your concerns.

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+   * based on the provided <code>properties</code> bundle.
+   * Logs using the provided <code>logger</code>. If there is a type issue, or an access problem
+   * then a <code>ConfigurationException</code> will be thrown.
+   *
+   * @param configurable Any properties must be modifiable via setter methods.
+   * @param properties Map&lt;String, String&gt;
+   * @param logger org.slf4j.Logger
+   * @throws ConfigurationException
+   */
+  public static void setConfigurationFields(Object configurable, Map<String, String> properties,
```
adenes: Could you please remove the `logger` parameter of this method?
I think it's better to have a private logger as all the logged exceptions are propagated to the caller so they have the chance to log the errors with their own loggers.
Plus if a caller doesn't have a logger then it can't call this method. Or - which is even worse - can try to call it with `null` logger and then bump into unexpected NPExceptions.

flume-ng-core/src/test/java/org/apache/flume/source/http/FlumeHttpServletRequestWrapper.java
```diff
@@ -52,273 +63,282 @@ public FlumeHttpServletRequestWrapper(String data) throws UnsupportedEncodingExc
     this(data, "UTF-8");
   }
 
-  @Override
```
adenes: These `@Override`s are not needed to be removed as these methods still override the ones from the parent class. Could you please put them back + also add them to the newly added methods? Thanks.

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+          Logger logger) throws ConfigurationException {
+    Class<?> clazz = configurable.getClass();
+
+    for (Method method : clazz.getMethods()) {
+      String methodName = method.getName();
+      if (methodName.startsWith("set") && method.getParameterTypes().length == 1) {
+        String fieldName = methodName.substring(3);
```
mpercy, line 55: If this is to be case sensitive, it should probably lowercase the first letter so that the property fooBar maps to setFooBar() which would be consistent with most of the rest of Flume.

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
mpercy: Perhaps mark as @InterfaceAudience.Private and @InterfaceStability.Evolving for a release

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+            } else if (fieldType.equals(double.class)) {
+              method.invoke(configurable, Double.parseDouble(value));
+            } else if (fieldType.equals(char.class)) {
+              method.invoke(configurable, value.charAt(0));
+            } else if (fieldType.equals(byte.class)) {
+              method.invoke(configurable, Byte.parseByte(value));
+            } else if (fieldType.equals(String[].class)) {
+              method.invoke(configurable, (Object)value.split("\\s+"));
+            } else {
+              logger.error("Unable to set " + fieldName + " as it is an unsupported type: "
```
mpercy: Should this just throw?

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+   * @param configurable Any properties must be modifiable via setter methods.
+   * @param properties Map&lt;String, String&gt;
+   * @param logger org.slf4j.Logger
+   * @throws ConfigurationException
+   */
+  public static void setConfigurationFields(Object configurable, Map<String, String> properties,
+          Logger logger) throws ConfigurationException {
+    Class<?> clazz = configurable.getClass();
+
+    for (Method method : clazz.getMethods()) {
```
mpercy: Should we do this the opposite way, where we try to call a setter for each property in the map, or that isn't going to be viable?

flume-ng-core/src/test/java/org/apache/flume/source/http/FlumeHttpServletRequestWrapper.java
```diff
@@ -52,273 +63,282 @@ public FlumeHttpServletRequestWrapper(String data) throws UnsupportedEncodingExc
...
   public int getRemotePort() {
     throw new UnsupportedOperationException("Not supported yet.");
   }
 
-  @Override
   public String getLocalName() {
     throw new UnsupportedOperationException("Not supported yet.");
   }
 
-  @Override
```
mpercy: Why remove all these `@Override`s? There is no longer an inheritance relationship?

flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -1719,25 +1719,42 @@ unavailable status.
...
 interceptors.*
-enableSSL          false                                         Set the property true, to enable SSL. *HTTP Source does not support SSLv3.*
-excludeProtocols   SSLv3                                         Space-separated list of SSL/TLS protocols to exclude. SSLv3 is always excluded.
-keystore                                                         Location of the keystore includng keystore file name
-keystorePassword                                                 Keystore password
+enableSSL             false                                         Set the property true, to enable SSL. *HTTP Source does not support SSLv3.*
+excludeProtocols      SSLv3                                         Space-separated list of SSL/TLS protocols to exclude. SSLv3 is always excluded.
+keystore                                                            Location of the keystore includng keystore file name
+keystorePassword                                                    Keystore password
+QueuedThreadPool.*                                                  Jetty specific settings to be set on org.eclipse.jetty.util.thread.QueuedThreadPool.
```
mpercy, line 1747: This is clearly a highly flexible approach, but do we really want to adopt all public methods of Jetty being configurable via Flume? Flume has traditionally been nearly 100% backwards compatible and if we directly expose Jetty public methods to users then are we writing checks we can't cash later on? Say if Jetty changes some method names in a future major version that we would like to upgrade to.

adenes (changes requested):
Thank you @tmgstevens  for the patch, I have added some comments to it. Let me know if something is not clear. Thanks.

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
tmgstevens: Would be nicer, but we'd need to infer the type because there isn't a Class.getMethod(String name) method.

flume-ng-doc/sphinx/FlumeUserGuide.rst
tmgstevens: Note the concern, but we've got plenty of users requesting additional configuration. It feels wrong to copy the params in one-by-one as we'll forever be behind the curve. Do you have any thoughts as to how we can improve the configurability? 

I can only think of shading the three relevant classes in a test and specifically including a JUnit to test that properties don't disappear.

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
tmgstevens: Done

flume-ng-core/src/test/java/org/apache/flume/source/http/FlumeHttpServletRequestWrapper.java
tmgstevens: Done

flume-ng-core/src/main/java/org/apache/flume/tools/FlumeBeanConfigurator.java
tmgstevens: Done

adenes:
Thanks @tmgstevens for the updated patch. The change looks good to me, but I'd like to hear the opinion of other reviewers as well. @szaboferee @mpercy

szaboferee (approved):
The changes look good.
Tests ran successfully

adenes (approved):
Thanks @tmgstevens for the patch and all the others for the review. I have added a minor change to the code before committing because the user guide failed to compile properly.

PR #159, opened 2017-08-26
FLUME-3149 reduce cpu cost for taildir file source while still maintaining reliability by using posFile in memory channel

zyfo2:
File channel tracks transferred events and use transnational mechanism to make transfer recoverable. However, it increases CPU cost due to frequent system calls like write, read, etc. The Cpu cost could be very high if the transfer rate is high. 
In contrast, Memory channel has no such issue which requires only about 10% of CPU cost in the same environment but it's not recovered if the system is down accidentally.
For sources like taildir, I propose we could write position file in memory channel to achieve reliability and reduce CPU cost.
After testing on my own production environment, CPU usage dropped from 13% to 3% and still maintain reliability.  （Transfer rate: 1Mb/s ,  kafka sink, file channel -> memory channel with pos file)

### DIFF
--- a/flume-ng-doc/sphinx/FlumeUserGuide.rst
+++ b/flume-ng-doc/sphinx/FlumeUserGuide.rst
@@ -1182,7 +1182,6 @@ headers.<filegroupName>.<headerKey> --                             Header value 
 byteOffsetHeader                    false                          Whether to add the byte offset of a tailed line to a header called 'byteoffset'.
 skipToEnd                           false                          Whether to skip the position to EOF in the case of files not written on the position file.
 idleTimeout                         120000                         Time (ms) to close inactive files. If the closed file is appended new lines to, this source will automatically re-open it.
-writePosInterval                    3000                           Interval time (ms) to write the last position of each file on the position file.
 batchSize                           100                            Max number of lines to read and send to the channel at a time. Using the default is usually fine.
 backoffSleepIncrement               1000                           The increment for time delay before reattempting to poll for new data, when the last attempt did not find any new data.
 maxBackoffSleep                     5000                           The max time delay between each reattempt to poll for new data, when the last attempt did not find any new data.
--- a/flume-ng-doc/sphinx/FlumeUserGuide.rst
+++ b/flume-ng-doc/sphinx/FlumeUserGuide.rst
@@ -1161,7 +1161,7 @@ If the new lines are being written, this source will retry reading them in wait 
 This source is reliable and will not miss data even when the tailing files rotate.
 By default it periodically writes the last read position of each files on the given position file in JSON format.
 It can also be configured by setting writePosOnCommit to true and use MemoryChannel to write the last consumed position of each file when commits.
-This way, we save CPU costs in FileChannel and maintains reliability as well.
+This way, we reach reliability and avoid heavy system calls in FileChannel thus reducing CPU cost.
 If Flume is stopped or down for some reason, it can restart tailing from the position written on the existing position file.
 
 In other use case, this source can also start tailing from the arbitrary position for each files using the given position file.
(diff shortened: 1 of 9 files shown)


flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -1182,7 +1182,6 @@ headers.<filegroupName>.<headerKey> --                             Header value
 byteOffsetHeader                    false                          Whether to add the byte offset of a tailed line to a header called 'byteoffset'.
 skipToEnd                           false                          Whether to skip the position to EOF in the case of files not written on the position file.
 idleTimeout                         120000                         Time (ms) to close inactive files. If the closed file is appended new lines to, this source will automatically re-open it.
-writePosInterval                    3000                           Interval time (ms) to write the last position of each file on the position file.
```
bessbd: Do we have to remove this?

flume-ng-sdk/src/main/java/org/apache/flume/event/SimpleEvent.java
```diff
@@ -64,4 +64,7 @@ public String toString() {
     return "[Event headers = " + headers + ", body.length = " + bodyLen + " ]";
   }
 
+  public void notifySource() {}
```
bessbd, line 71: Can you please add javadoc to clarify usage?

flume-ng-sdk/src/main/java/org/apache/flume/event/SimpleEvent.java
```diff
@@ -64,4 +64,7 @@ public String toString() {
...
+
+  public void commit() {}
```
bessbd, line 79: Can you please add javadoc to clarify usage?

flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/TaildirSource.java
```diff
@@ -104,16 +107,15 @@ public synchronized void start() {
     } catch (IOException e) {
       throw new FlumeException("Error instantiating ReliableTaildirEventReader", e);
     }
+    for (Entry<Long, TailFile> entry : reader.getTailFiles().entrySet()) {
+      inodePositionMap.put(entry.getKey() + "", entry.getValue().getPos() + "");
```
bessbd: Are the `+ ""`-s needed? As per https://stackoverflow.com/a/1572724/5323166 , `String.valueOf` is recommended. Please, let me know if the intent was different.

bessbd (commented):
Overall, the changes seem to be fine. Can you please provide a way of testing this change? Ie. "run this command to see the performance gain".

flume-ng-sdk/src/main/java/org/apache/flume/event/SimpleEvent.java
zyfo2: sure.

flume-ng-sdk/src/main/java/org/apache/flume/event/SimpleEvent.java
zyfo2: sure.

flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/TaildirSource.java
zyfo2: Thanks for pointing out. I'll fix it.

flume-ng-doc/sphinx/FlumeUserGuide.rst
zyfo2: Perhaps it's better to keep it to be compatible with other channels. I'll add it back and modify the user guide.

zyfo2:
I think this change is essentially reaching reliability in MemoryChannel when using taildirSource. So the performance diff is essentially the diff of MemoryChannel  and FileChannel. In my tests it saves more than 90% percent of CPU but it's no easy way of comparing it by a simple command.
Perhaps I should add some tests to verify the reliability when using tailDirSource and MemoryChannel?

asfgit:

PR #161, opened 2017-08-28
FLUME-3155: Use batch mode in mvn to fix Travis CI error

mcsanady:
Reduce log verbosity with mvn batch mode to avoid this error:
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated


szaboferee (approved):
LGTM :+1:

PR #162, opened 2017-08-28
FLUME-3148 Make tests run in parallel

szaboferee:
updating surefire plugin, setting fork options, adding a testPortProvider

Refactoring tests to use a common TestPortProvider
Splitting Kafka test to smaller test classes avoiding timeout

turning on parallel test running


simonati (changes requested):
Hi @szaboferee,

The diff passes {{mvn clean install}} on on my side, and only takes roughly half of the original time to be executed (25min vs 50min) so at first glance this looks like a killer change! Thank you for that!

My problem is that it is large to review it in a single go. Would you mind splitting it up into more consumable chunks. eg: yaml changes, consolidation of surefire versions, setting up parallel test execution, introduction of the new shared module, fixed server port allocation, derby change to use memory instead of file, expansion of kafka tests into smaller classes, etc...
Let me know if I can help you with this.

szaboferee:
Hi @simonati,

thanks for the review! sure, I will make more consumable tickets and pull requests.
Closing this one.

PR #163, opened 2017-09-01
FLUME-3127 Upgrade libfb303 library dependency

szaboferee:
- thrift version changed to 0.9.3
- thrift objects generated with the new version
- deprecated method calls updated in ScribeSource

Full build with tests ran successfuly


szaboferee:
I ran a test with flume 1.6 and 1.7 as an agent that had a thrift sink connected to an agent running with the new thrift version and the events were transferred successfully

adenes:
Thank you @szaboferee for the patch, it seems good to me.
The code compiles, tests pass. Thanks for testing the backward compatibility.
👍 
I'm going to commit this change.

PR #164, opened 2017-09-01
FLUME-3157 Refactor TestHDFSEventSinkOnMiniCluster to not use LeaseMa…

szaboferee:
…nager private API

test refactored, it has the same result and behaviour

Test ran successfully after the refactor


adenes (changes requested):
Hi @szaboferee,
Thank you for the patch.
Could you please make it a bit more clear by adding a comment that the `NameNodeAdapter.getLeaseRenewalTime` returns `-1` if and only if there is no lease for the given path?
Thank you.

adenes (approved):
Thanks @szaboferee for the update.
Tests pass with your patch in place, but do fail without the fix introduced in FLUME-3080 / PR #127 
👍

adenes:
I might haven't been clear enough in my previous comment:
I meant that I verified that the test behaves as before your change.
- currently it passes, which is good
- but removing the change which the affected test cases were written for (i.e. #127, commenting out the body of `BucketWriter.recoverLease()`) they fail as expected, so they still test what they are supposed to test
-> the behaviour didn't change indeed.

PR #166, opened 2017-09-03
[SMALLFIX]fix an infinite recursion

TeddyBear1314:
I found this method has never been called in flume, but it is a hidden bug, I think.


adenes (approved):
Good catch, thank you @TeddyBear1314 
+1 from me

PR #167, opened 2017-09-05
modifying log4j file read repeatedly error

HbnKing:
The  `log4j `files always segmentalized  the file by date and  log file size . but the  `taildirSource` regard it as a new file    .so the  data collected repeatly.


szaboferee:
Hi @HbnKing  
Thank you for the contribution!
 I was wondering if there is a JIRA ticket for this issue that can be mentioned in the title? 

Is this problem reproducible with a unit/integration test? 
If it is, could you also write such a test so we could understand the fix better?

asfgit:

PR #168, opened 2017-09-11
FLUME-3175. Turn off doclint for javadoc generation.

adenes:
With Java 8's new doclint feature the javadoc generation breaks due to the
strict checking so adding -Xdoclint:none to the maven-javadoc-plugin's
configuration to turn off the check.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -519,6 +519,9 @@ limitations under the License.
                 </goals>
               </execution>
             </executions>
+            <configuration>
+              <additionalparam>-Xdoclint:none</additionalparam>
+            </configuration>
           </plugin>
 
           <plugin>


bessbd:
Hi Denes,

Right now I don't have a Flume dev environment available. Do you have an estimation on how much of an effort it would be to fix the javadoc issues?


Thanks,

Donat

adenes:
Hi @bessbd,
I've created a gist with the error messages without the doclint turned off: https://gist.github.com/adenes/d1ebcf9f5ff3d521bb0ed492017f9288
~600 lines, 3 lines/error -> ~200 errors

Not on insanely huge number but it'd take quite a lot of effort to fix all those, that's why I decided to turn the checks off. I filed a Jira ticket to fix the issues: https://issues.apache.org/jira/browse/FLUME-3176

szaboferee (approved):
Hi @adenes it looks good to me. Tested the build without it and it reported the errors.  So the fix works.
Plus one for the jira for fixing them.

PR #169, opened 2017-09-12
FLUME-3173: Upgrade joda-time

mcsanady:
Unit tests are passing.
Found no suspicious change in [joda-time release doc.](http://www.joda.org/joda-time/changes-report.html)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@ limitations under the License.
     <javax-jms.version>1.1-rev-1</javax-jms.version>
     <jersey.version>1.8</jersey.version>
     <jetty.version>6.1.26</jetty.version>
-    <joda-time.version>2.1</joda-time.version>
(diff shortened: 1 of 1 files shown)


marcellhegedus:
I think we can go with the latest stable release (v.2.9.9) which has the current ([2017b](http://www.iana.org/time-zones)) tz information out of the box.

mcsanady:
@marcellhegedus I upgraded to v 2.9.6, because you tested s3 authentication with this version. Could you test this with v2.9.9 too?

marcellhegedus:
@mcsanady Right. Retested s3 authentication and it also works with joda-time-2.9.9.jar. Thanks for the patch! LGTM +1

adenes:
Thank you @mcsanady for the patch and @marcellhegedus for the review, I'm about to commit this.

PR #170, opened 2017-09-12
Fix NetCat UDP Source table in FlumeUserGuide.rst

adenes:
Netcat UDP source's table in the user guide was broken, rendering failed with _Malformed table. Text in column margin at line offset 7_ error.


szaboferee:
Trivial fix, looks good to me :+1:

PR #171, opened 2017-09-12
Set the copyright date dynamically in documentation footer

adenes:
Display "2009-current year" instead of the hardcoded "2009-2012".


szaboferee (approved):
Tested it with
mvn clean site -P site 
The result contained the proper copyright information.
:+1:

PR #173, opened 2017-09-14
FLUME-3179 Fix file channel size

movit:


szaboferee (commented):
Hi @movit 

Thank you for your contribution!

Is it possible to add a test case that tests this fix so we could make sure that another change in the future would not reintroduce this bug?

asfgit:

PR #175, opened 2017-09-15
Fix source assembly

adenes:
flume-checkstyle breaks the assembly because its parent is not the flume-parent
Removing the moduleSets definition from the src assembly solved the issue.
Files are added based on fileSets, the resulting tarball's content equals
to the result of the dev-support/generate-source-release.sh in a clean
working directory.

### DIFF
--- a/flume-ng-dist/src/main/assembly/src.xml
+++ b/flume-ng-dist/src/main/assembly/src.xml
@@ -43,6 +43,7 @@
         <exclude>**/*.iml</exclude>
         <exclude>**/.settings/**</exclude>
         <exclude>lib/**</exclude>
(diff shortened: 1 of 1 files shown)


szaboferee (commented):
I would .DS_Store to the exclude list

mcsanady (approved):
The build success confirmed.

adenes:
Thanks @szaboferee , I have added the .DS_Store to the excludes

szaboferee (approved):
Looks good to me. 
I ran the build and checked the files. They were ok.

mpercy:
Did you check to make sure that checkstyle still works?

PR #177, opened 2017-10-01
FLUME-3159 setting test connection to in-memory derby to make paralle…

szaboferee:
…l test run possible


adenes (approved):
Thank you @szaboferee for the patch, the change looks good to me, +1

PR #178, opened 2017-10-18
FLUME-3186: Make asyncHbaseClient configuration parameters available …

mcsanady:
In 1.7 version AsyncHbaseClient it is possible to pass all configuration to the constructor.
This makes it possible to pass all the parameters to this library with a specific prefix.

### DIFF
--- a/flume-ng-sinks/flume-ng-hbase-sink/src/main/java/org/apache/flume/sink/hbase/AsyncHBaseSink.java
+++ b/flume-ng-sinks/flume-ng-hbase-sink/src/main/java/org/apache/flume/sink/hbase/AsyncHBaseSink.java
@@ -109,7 +109,7 @@ public class AsyncHBaseSink extends AbstractSink implements Configurable {
   private long batchSize;
   private static final Logger logger = LoggerFactory.getLogger(AsyncHBaseSink.class);
   private AsyncHbaseEventSerializer serializer;
-  private Config asyncClientConfig;
+  Config asyncClientConfig;
   private String eventSerializerType;
   private Context serializerContext;
   private HBaseClient client;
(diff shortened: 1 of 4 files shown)


mcsanady:
Tests:
I have uploaded the binary to a cluster, where I could use a configuration with values like this:
agent1.sinks.asynchbase.async.hbase.region_client.inflight_limit = 2

flume-ng-sinks/flume-ng-hbase-sink/src/main/java/org/apache/flume/sink/hbase/AsyncHBaseSink.java
```diff
@@ -422,6 +425,19 @@ public void configure(Context context) {
         context.getInteger(HBaseSinkConfigurationConstants.CONFIG_MAX_CONSECUTIVE_FAILS,
                            HBaseSinkConfigurationConstants.DEFAULT_MAX_CONSECUTIVE_FAILS);
 
+
+    ImmutableMap<String, String> asyncProperties
```
adenes: Could you please use standard `Map` instead of the `ImmutableMap`?

adenes (changes requested):
Thanks @mcsanady for the patch.
In general I'm ok with the change, I had one inline comment + the following two:
- could you please create some unit tests for the change? You could verify whether the expected key-value pairs get into the `asyncClientConfig` instance.
- Please update the User Guide to reflect the introduced changes. Don't forget to mention what is the precedence of the "old" and the "new" way of setting the `zookeeperQuorum` and `znodeParent` parameters.

Thanks!

szaboferee (commented):
Thank You for the contribution, this feature is very useful!
Because of it's usefulness I would love to see it documented in the manual so others could use it as adenes already said.
 
Otherwise it looks good so far.

adenes (approved):
Thanks for the update.
Tests pass, the user guide looks good.
+1

PR #181, opened 2017-11-03
FLUME-2050 Upgrade to Log4j 2

rgoers:
This PR removes the use of log4j 1.x from Flume.


flume-ng-sinks/flume-hive-sink/metastore_db/README_DO_NOT_TOUCH_FILES.txt
```diff
@@ -0,0 +1,9 @@
+
```
szaboferee: Hi @rgoers , thanks for the pull request
I am wondering how this metastore_db folder is necessary to upgrade the log4j version.

szaboferee (changes requested):

flume-ng-sinks/flume-hive-sink/metastore_db/README_DO_NOT_TOUCH_FILES.txt
szaboferee: This is some leftover from a failed test run. Could you please remove it so we could review the real changes?

rgoers: I didn't modify those files so I'm not sure how they got in the commit. Out of curiosity, is there a reason they aren't copied to the target directory if the tests are going to modify them? Maven convention is that only items in the target directory should be modified by the build so I don't have my gitignore set up to skip these files.

rgoers: I will revert those files as soon as I can.

rgoers:
The extra files have been removed.

szaboferee:
@rgoers 
As the first test, I have started a simple test agent with the trunk and with the patch and I see some difference between the log format
before:
`07 nov. 2017 14:25:00,226 INFO  [conf-file-poller-0] (org.apache.flume.conf.FlumeConfiguration$AgentConfiguration.addProperty:930)  - Added sinks: k1 Agent: a1`

after:
`07 Nov 2017 14:25:06,189 INFO  [conf-file-poller-0] (org.apache.flume.conf.FlumeConfiguration$AgentConfiguration.addProperty:930) [] - Added sinks: k1 Agent: a1`

I believe this upgrade should not introduce any change in the default log format.

rgoers:
The format strings are exactly the same. The difference seems to be how %x is handled. In Log4j 1.x the NDC seems to be returned as null if there is nothing, so I am not sure why it is an empty string. In Log4j 2 the ThreadContext Stack returns an empty list which I believe is why you see "[]".  I am curious - Does Flume even use the NDC? Very view applications use the NDC - most use the MDC/ThreadContext Map. A grep of Flume finds the only occurrence of NDC is in the Log4jAppender and that reference only retrieves the NDC from the logging event.

Log4j 2 has a way to handle this so I will give it a try and update the PR when I confirm it is working.

rgoers:
@szaboferee I ran the agent in flume-ng-dist and the results I get don't match yours. With log4j 1.x I got
``07 Nov 2017 23:28:57,156 INFO  [lifecycleSupervisor-1-0] (org.apache.flume.node.PollingPropertiesFileConfigurationProvider.start:62)  - Configuration provider starting``
This doesn't match what you have above. For some reason you have the month as "nov.". I am not sure how you are getting that as the pattern uses "MMM" which is handled by SimpleDateFormat in Log4j 1, which prints Nov as I have here.

Also, based on what you posted above I modified the log4j 2 configuration so that it doesn't print anything if there is no NDC and there is only one space as that is what I see above. However, when I look at the output I get from Log4j 1 I am seeing 2 spaces before the '-'. I noticed that when I include that string here that the double space is being converted to a single space, so I am guessing I really need to modify the pattern to emit 2 spaces when the NDC isn't present instead of just 1.

Are you using the standard conf/log4j.properties in the gzipped tar built by flume-ng-dist? If you are I don't know how you are getting "nov.".

conf/log4j.properties
```diff
@@ -1,68 +0,0 @@
-#
```
szaboferee, line 1: I was wondering why were you changing log4j.properties to XML.
Are there benefits I am not aware of?

There might be systems where flume was integrated with log4j.properties format. What is the precedence if there are both .properties and .xml?

flume-ng-clients/flume-ng-log4jappender/pom.xml
```diff
@@ -67,6 +67,7 @@ limitations under the License.
     <dependency>
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
+      <version>1.2.17</version>
```
szaboferee, line 70: Wouldn't be better to have this in the parent dependencyManagement as there are transitive dependencies that are still using log4j 1.x?

flume-ng-sinks/flume-http-sink/src/main/java/org/apache/flume/sink/http/HttpSink.java
```diff
@@ -27,7 +27,8 @@
 import org.apache.flume.conf.Configurable;
 import org.apache.flume.instrumentation.SinkCounter;
 import org.apache.flume.sink.AbstractSink;
-import org.apache.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
```
szaboferee, line 30: I would prefer the slf4j logger here because that way the user could decide to use a different logger implementation.
(despite the fact that the original code already used log4j)

conf/log4j.properties
rgoers: Log4j 2 doesn't support the log4j 1 properties format. Even if the log4j 2 properties format is used it is still different than the log4j 1 format. I prefer the XML format.

flume-ng-clients/flume-ng-log4jappender/pom.xml
rgoers: No. All references to log4j 1 have been removed except for this one instance.

flume-ng-sinks/flume-http-sink/src/main/java/org/apache/flume/sink/http/HttpSink.java
rgoers: Because it was using log4j I converted it to use log4j...

flume-ng-clients/flume-ng-log4jappender/pom.xml
szaboferee: All references in flume however, some dependencies of flume has it as a transitive dependency.
For example hive-shmis, hadoop-common, kafka_2.10.

We should harmonize the versions for these dependencies as well to avoid version mismatches.

conf/log4j.properties
szaboferee: What happens if there is a log4j.properties (log4j 2 format) and a log4j.xml file on the classpath at the same time?

flume-ng-sinks/flume-http-sink/src/main/java/org/apache/flume/sink/http/HttpSink.java
szaboferee: Flume components are using slf4j. I guess this log4j logger in the HttpSink was overlooked at the review and that is why it is not an slf4j logger. It would be nice if it could be consistent.

conf/log4j.properties
rgoers: Log4j uses an @Order annotation on the ConfigurationFactories to determine which factories get precedence. Properties have the highest, followed by yaml and then XML. So if a properties file is present it will be used instead of the xml file.

flume-ng-clients/flume-ng-log4jappender/pom.xml
rgoers: I went through and excluded all the transitive dependencies to log4j in the Flume pom.xml files that I could find. They have been replaced with references to log4j-1.2-api.

flume-ng-sinks/flume-http-sink/src/main/java/org/apache/flume/sink/http/HttpSink.java
rgoers: I have no problem if that change is made but it doesn't necessarily have to happen as part of this PR.

szaboferee: I agree. I will make a different PR about it. Thank you.

flume-ng-clients/flume-ng-log4jappender/pom.xml
szaboferee: Seeing the output of `mvn dependency:tree` there are a few places where it looks like you might have missed it. Could you please double check it to see if I am right? 

Wouldn't it be easier to use the dependencyManagement in parent pom to exclude the old version?

conf/log4j.properties
szaboferee: Great, thanks for the information.

flume-ng-clients/flume-ng-log4jappender/pom.xml
rgoers: It might. I will take a look again.

rgoers:
I have updated the PR again but it can't be merged. It requires Log4j 2.10, which hasn't been released yet, to resolve a ClassNotFound error in Hadoop. The release process for Log4j 2.10 will be starting in the next few days. I will update this PR when it has completed.

rgoers:
The patch has been upgraded to use Log4j 2.10.0. That version includes a version of AppenderSkeleton from Log4j 1 which should eliminate problems with Hadoop. From what I can tell this PR should be ready to go now.

szaboferee (approved):
I checked the dependency tree, the only log4j 1.x is in the appender, tests ran successfully.

The change looks good to me.

PR #182, opened 2017-11-13
FLUME-3158: Upgrade surefire version and config

mcsanady:
Set surefire version to the latest and configure it to rerun failed tests. This way the flaky tests will not break the build as often.

Tests I did: building Flume locally, and running unit tests on server environments.

### DIFF
--- a/flume-ng-sinks/flume-ng-morphline-solr-sink/pom.xml
+++ b/flume-ng-sinks/flume-ng-morphline-solr-sink/pom.xml
@@ -103,13 +103,9 @@ limitations under the License.
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>${surefire.version}</version>
         <configuration>
           <argLine>-Dtests.locale=en_us</argLine>
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
-          <systemPropertyVariables>
-            <!--<solr.expected.version>${solr.expected.version}</solr.expected.version>-->
-          </systemPropertyVariables>
         </configuration>
       </plugin>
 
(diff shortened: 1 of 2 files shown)


flume-ng-sinks/flume-ng-morphline-solr-sink/pom.xml
```diff
@@ -32,7 +32,7 @@ limitations under the License.
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <solr.version>${solr-global.version}</solr.version>
     <solr.expected.version>${solr-global.version}</solr.expected.version> <!-- sanity check to verify we actually run against the expected version rather than some outdated version -->
-    <surefire.version>${mvn-surefire-plugin-old-morphline.version}</surefire.version>
+    <surefire.version>${mvn-surefire-plugin.version}</surefire.version>
```
szaboferee: I would remove this property and also the version tag from the part:
`        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>${surefire.version}</version>`

so the plugin could use the pluginManagement settings from the parent pom

szaboferee (changes requested):

flume-ng-sinks/flume-ng-morphline-solr-sink/pom.xml
mcsanady: `<plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <configuration>
          <argLine>-Dtests.locale=en_us</argLine>
          <redirectTestOutputToFile>true</redirectTestOutputToFile>
        </configuration>
      </plugin>`

szaboferee (approved):
Thanks for the upgrade! This will help to avoid build failures because of flaky tests.

The tests ran successfully, the change looks good to me.

PR #188, opened 2017-12-05
FLUME-3201 Fix SyslogUtil to handle RFC3164 format in december correctly

szaboferee:


tmgstevens:
LGTM. Sorry, my bug that one. I was just about to raise a PR for the fix myself!

tmgstevens:
@szaboferee Could I get your thoughts on https://github.com/tmgstevens/flume/commit/86390d6d0b2e9d0c3d89b5e76275afcbc3672d1c please?

I'd like to avoid us hitting this again, so I've modified SyslogUtils to use a package-private injectable clock and then changed the test so that it "runs" once for each of the next 13 months.

szaboferee:
@tmgstevens  I updated this pull request based on your code. Modified a little bit to avoid the static clock and the mutability. Also found  and fixed a small bug in the test at the monthOffset

mcsanady (approved):

flume-ng-core/src/test/java/org/apache/flume/source/TestSyslogUtils.java
```diff
@@ -196,41 +199,53 @@ public void TestRfc3164HeaderApacheLogWithNulls() throws ParseException {
...
+    for (int monthOffset = 0; monthOffset <= 13; monthOffset++) {
+      Clock mockClock = Clock.fixed(
+              LocalDateTime.now().plusMonths(monthOffset).toInstant(ZoneOffset.UTC),
+              Clock.systemDefaultZone().getZone()
+      );
+
+      //We're then going to try input dates (without the year) for all 12 months, starting
+      //10 months ago, and finishing next month (all relative to our mocked clock)
+      for (int i = -10; i <= 1; i++) {
+        SimpleDateFormat sdf = new SimpleDateFormat("MMM  d hh:MM:ss", Locale.ENGLISH);
```
vegiapparao, line 212: @szaboferee @mcsanady - I think the date format should be changed to `MMM dd hh:mm:ss`? I left a comment for @tmgstevens in my PR but haven't heard back from him yet.

szaboferee: @vegiapparao  good point.  Now minutes are referenced as months in the pattern. Luckily it is in  a test, so no production is affected. We should address that in a following pull request backed up with a jira ticket:.

tmgstevens: Agreed, although I'm now confused as to why the test passes.

vegiapparao: @tmgstevens, @szaboferee,  I have changed it locally, ran the full suite and all the tests are passing.  Shall I submit a patch for this?

tmgstevens: @vegiapparao okay, I've finally got my head around this...! SimpleDateFormat sdf is only used to format the desired date for the message. The desired date is then tested against the same date string with yyyy prepended, to make sure the the year has been inferred correctly. As such, the minute actually doesn't matter. It's useful to tidy up, but has no functional impact. Up to you if you want to raise a JIRA and submit a PR.

PR #192, opened 2017-12-26
FLUME-3205:Remove unnecessary 'synchronized' in ResettableFileInputStream

tcsdn:
I found no concurrent access to ResettableFileInputStream object in flume code.
Remove 'synchronized' will improve performance of SpoolDirectorySource.
Tested with a file of 200MB，with 'synchronized' it took 9 seconds to read the file.
Without 'synchronized' it took 7 seconds.


tcsdn:
@harishreedharan 
Can you help me review the change?

bessbd:
Hi @Guangxian ,

Thank you for the PR!

I've taken a look at the change and it seems to be a nice catch!

Even though there might be no concurrent access to the method in the Flume codebase, there might be access "from the outside". Thus, it might be a good idea to make this change in a backwards compatible way. (Ie. extract the "contents" of the method and leave a synchronized read that calls the non-synchronized, new one or something like that.)

Please, let me know what you think.


Thank you,

Donat

tcsdn:
@bessbd Thank you for review. I add a `NonSyncResettableFileInputStream` class. Let `ResettableFileInputStream` extends `NonSyncResettableFileInputStream`. Please review it.

asfgit:

PR #195, opened 2018-01-26
FLUME-2957. Remove Guava from our public API

adenes:
The only Guava class used in public API is the `ImmutableMap` in the `org.apache.flume.Context`, `org.apache.flume.node.MaterializedConfiguration` and `org.apache.flume.node.SimpleMaterializedConfiguration` classes.

This commit replaces these with `java.util.Map`.

Note: this is a breaking change as it modifies the return type of a method on a public interface.

### DIFF
--- a/flume-ng-configuration/src/main/java/org/apache/flume/Context.java
+++ b/flume-ng-configuration/src/main/java/org/apache/flume/Context.java
@@ -48,8 +48,9 @@ public class Context {
    * Gets a copy of the backing map structure.
    * @return immutable copy of backing map structure
    */
-  public ImmutableMap<String, String> getParameters() {
+  public Map<String, String> getParameters() {
     synchronized (parameters) {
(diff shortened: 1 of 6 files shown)


mpercy:
Hi Denes, this patch looks good to me. However this is a breaking API change. Would you mind starting a conversation on the dev list about potentially planning for a Flume 2.0.0 to support this backwards incompatible change?

szaboferee (approved):
I have checked for other possible guava classes in return types or parameters and found
`com.google.common.eventbus.EventBus` in `PollingPropertiesFileConfigurationProvider`.
We might not consider it as a public API, so it can be safe to leave it there.

Otherwise, it looks good to me.

mpercy (approved):
+1 let's commit this. It needs a cherry-pick / squash after the force-push to trunk.

PR #197, opened 2018-02-13
FLUME-2442 Need an alternative to providing clear text passwords in flume config

szaboferee:
Adding an interface with 3 implementations to provide functionality at the configuration level to replace variables/keys from external sources. This component is capable of hiding sensitive information or injecting generated data into the configuration. 

The implementation affects only the configuration layer so existing components (sinks/sources/channels/etc) does not have to change and new components can already have it through the configuration.  

New custom implementations can be easily added even in plugin form. 

Each implementation has unit tests in their module and an integration test in the flume-ng-tests module.

Documentation starts [here](https://github.com/szaboferee/flume/blob/fa13593baa06c9d770a21fc970110e0c9abf2ef8/flume-ng-doc/sphinx/FlumeUserGuide.rst#configuration-filters)


flume-ng-configfilters/flume-ng-external-process-config-filter/src/main/java/org/apache/flume/configfilter/ExternalProcessConfigFilter.java
```diff
@@ -0,0 +1,98 @@
...
+              this.command, p.exitValue(), stderr));
+    }
+
+
+    return getResultFromStream(p.getInputStream());
+  }
+
+  private String getResultFromStream(InputStream inputStream) {
+    try (Scanner scanner = new Scanner(inputStream, charset.name())) {
+      return scanner.useDelimiter("\\A").next().trim();
```
adenes: Why is the `useDelimiter("\\A")` needed here?
Or why not use the good old `BufferedReader.readLine()`?

flume-ng-configuration/src/main/java/org/apache/flume/conf/FlumeConfiguration.java
```diff
@@ -348,75 +381,135 @@ private boolean isValid() {
...
+        //JAVA EL expression style ${myFilterName[my_key]}
+        List<String> delimiters = Arrays.asList("'", "\"", "");
+        for (String delimiter: delimiters) {
+          String prefix = "${" + configFilter.getName() + "[" + delimiter;
+          String suffix = delimiter + "]}";
+          boolean matchesWithDelimiter = value.startsWith(prefix) && value.endsWith(suffix);
+
+          if (matchesWithDelimiter) {
+            String filteredValue = configFilter.filter(value.substring(
+                configFilter.getName().length() + delimiter.length() + 3,
```
adenes: This can be simplified to `prefix.length()`

adenes: BTW, why not regexp with capture group and back reference for the delimiter?

flume-ng-configuration/src/main/java/org/apache/flume/conf/FlumeConfiguration.java
```diff
@@ -348,75 +381,135 @@ private boolean isValid() {
...
+        //JAVA EL expression style ${myFilterName["my_key"]} or
+        //JAVA EL expression style ${myFilterName[my_key]}
+        List<String> delimiters = Arrays.asList("'", "\"", "");
+        for (String delimiter: delimiters) {
+          String prefix = "${" + configFilter.getName() + "[" + delimiter;
+          String suffix = delimiter + "]}";
+          boolean matchesWithDelimiter = value.startsWith(prefix) && value.endsWith(suffix);
+
+          if (matchesWithDelimiter) {
+            String filteredValue = configFilter.filter(value.substring(
```
adenes: Should we catch & log `Exception`s (or `Throwable`s) here instead of relying on the implementations? What'd happen if there is an unexpected exception in the `filter()` method?

adenes (commented):
Thank you @szaboferee for the patch, it's a really valuable contribution.

I had a look on your change and some questions came up immediately:
- doesn't this interfere with the already existing `EnvVarResolverProperties`?
- wouldn't it be better to have the default implementations in one module? Or maybe the implementations and the API (`ConfigFilter` and `AbstractConfigFilter`) classes as well? I don't have a strong opinion on this but having 1 class + 1 test in one module seems to be a bit of an overkill for me.

I also added some inline comments, but I'll continue reviewing your patch soon.

flume-ng-configfilters/flume-ng-external-process-config-filter/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+                                 <executable>/bin/sh</executable>
+                                 <arguments>
+                                     <argument>-c</argument>
+                                     <argument>
+                                         set -x
+
+                                         SRC="${basedir}/src/test/resources"
+                                         DST="${project.build.directory}/test-classes"
+
+                                         find "$$SRC" -printf "%P\0" | xargs --verbose -0 -I {} chmod --reference="$$SRC/{}" -f "$$DST/{}"
```
mcsanady: --verbose switch does not work on macos

mcsanady (changes requested):

flume-ng-configfilters/flume-ng-external-process-config-filter/src/main/java/org/apache/flume/configfilter/ExternalProcessConfigFilter.java
szaboferee: using the delimiter \A (begining of the string) lets scanner to read the whole input and not just one line. Scanner also has readLine();

I can change it to readLine if it makes more sense to just read the first line of the input.

flume-ng-configuration/src/main/java/org/apache/flume/conf/FlumeConfiguration.java
szaboferee: I will add a try catch here and the value will not be replaced if an exception has been thrown

flume-ng-configuration/src/main/java/org/apache/flume/conf/FlumeConfiguration.java
szaboferee: will change to regex and make it more flexible. I chose not to use regex, because it is always slower than simple string operations. Fortunately, it has to run only when config changes and only as many times as many lines the config has. A user has to have a large configuration to feel the difference, so It is ok to change it to regex.

flume-ng-configfilters/flume-ng-external-process-config-filter/pom.xml
szaboferee: thanks!

szaboferee:
@adenes thanks for the review!

EnvVarResolverProperties can still work with this, however, I find it better to have it with the same configuration method as the other config filters.

about the module structure:
the API module has to be separated if we would like to save the new custom implementation developers from depending on something they do not need. This is the proper dependency graph in my opinion.

the default implementations are already in a group module.
I have seen much less pain because of too many loosely coupled module than unnecessarily tightly coupled modules. 

If the API and the implementations would be in the same module, it would create the same kind of circular dependency that other implementations already have in the configuration module. Which I would really like to avoid.

adenes (approved):
Thank you for the explanation @szaboferee , I understand your points.
The change looks good to me, +1

mcsanady (approved):
LGTM

PR #200, opened 2018-03-08
Refactor ConfigurationProviders

andrasbeni:
Refactor ConfigurationProviders so

- they can be handled uniformly from Application
- they fit the push model better
- EventBus is unneccessary

Also  make TestPollingZooKeeperConfigurationProvider.EventSync.notified volatile

### DIFF
--- a/flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
+++ b/flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Consumer;
 
 import org.apache.flume.Channel;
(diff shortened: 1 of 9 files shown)


szaboferee (changes requested):
Hi @andrasbeni, thanks for the contribution.

the tests showed some error:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:testCompile (default-testCompile) on project flume-ng-node: Compilation failure`

`flume-ng/flume-ng-node/src/test/java/org/apache/flume/node/TestApplication.java:[171,8] error: constructor PollingPropertiesFileConfigurationProvider in class PollingPropertiesFileConfigurationProvider cannot be applied to given types;`

andrasbeni:
Hi @szaboferee,
It seems I did not base my change on the right branch. Before I continue let's clarify:
 - should I rebase on flume-1.x or trunk?
 - what is the java version to use.

szaboferee:
@andrasbeni  trunk and 1.8

PR #201, opened 2018-03-13
【乱码问题】characters are not displayed properly____execSource.java

xvjie:
I hope to confirm the following questions:
      When I use execSource to get data to elasticsearchSink:
      The source uses the "GBK" encoding, and the data on the es side will be garbled.
I think:
        When the exec source configuration code is "GBK", the data should be encoded as "UTF-8" after reading the data, because the elasticsearchSink side does not have a code configuration. So I made the following changes to execSource.java in my personal use:
        eventList.add(EventBuilder.withBody(line.getBytes("UTF-8")));
--------------------------------------------------------------------------------------------英文表达不清楚的话，以下是汉语原文
我在使用execSource获取数据到elasticsearchSink时：
source使用“GBK”编码，es端的数据会乱码。
我认为：
       exec源端配置编码为“GBK”时，在读取到数据以后应该将数据编码为"UTF-8"，因为elasticsearchSink端并没有编码配置。所以我在个人使用中将execSource.java作出以下修改：
       eventList.add(EventBuilder.withBody(line.getBytes("UTF-8")));

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
@@ -333,7 +333,7 @@ public class ExecSource extends AbstractSource implements EventDrivenSource, Con
           while ((line = reader.readLine()) != null) {
             sourceCounter.incrementEventReceivedCount();
             synchronized (eventList) {
-              eventList.add(EventBuilder.withBody(line.getBytes(charset)));
+              eventList.add(EventBuilder.withBody(line.getBytes("UTF-8")));
               if (eventList.size() >= bufferCount || timeout()) {
                 flushEventBatch(eventList);
               }


flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
```diff
@@ -333,7 +333,7 @@ public void run() {
           while ((line = reader.readLine()) != null) {
             sourceCounter.incrementEventReceivedCount();
             synchronized (eventList) {
-              eventList.add(EventBuilder.withBody(line.getBytes(charset)));
+              eventList.add(EventBuilder.withBody(line.getBytes("UTF-8")));
```
szaboferee, line 336: The `charset` variable you just removed contains the configurable charset which is `UTF-8` by default if you have not changed it in your configuration.

Did this change actually fix your issue?

szaboferee (changes requested):

flume-ng-core/src/main/java/org/apache/flume/source/ExecSource.java
xvjie: Yes, it is solved.
Look roughly at the data flow:
If I specify ‘a1.sources.r1.charset = GBK’ in the source configuration, the whole process of the data is encoded as follows:
ExecSource read data (GBK) -->> Data write eventList use encoding (GBK) --->> RegexExtractorInterceptor.java line 130 decoding (UTF-8, garbled) --->> elasticsearchSink default decoding (UTF-8 and no configuration items, Garbled)
So, I think that after reading out the data according to the configuration code, it should use the default UTF-8 encoding (getBytes()) instead of the configured encoding (GBK), otherwise the subsequent decoding (UTF- 8) Will be garbled.
----------------------------------------------------------------------原文，我的英语太糟糕
是的，解决了。
大致看下数据流程：
如果我在source配置中指定‘a1.sources.r1.charset = GBK’，那么数据整个过程编码如下：
execSource读数据（GBK）>> 数据写入eventList使用编码（GBK） >> RegexExtractorInterceptor.java第130行解码用（UTF-8，已乱码）>> elasticsearchSink默认解码（UTF-8且没有配置项，已乱码）
所以，我认为应该按照配置编码读取出数据后，应该使用使用默认的UTF-8来编码（getBytes（）），而不应该按照配置的编码（GBK）来编码，否则后续的解码（UTF-8）会乱码。

szaboferee: I see. Thanks for the clarification. In this case, the charset change should be either in the elasticsearchsink or should be a new config parameter like `targerCharset`

PR #202, opened 2018-03-13
FLUME-3223 Flume HDFS Sink should retry close prior to performing a r…

mcsanady:
…ecoverLease

### DIFF
--- a/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
+++ b/flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
@@ -114,6 +114,7 @@ class BucketWriter {
   // reopened. Not ideal, but avoids internals of owners
   protected boolean closed = false;
   AtomicInteger renameTries = new AtomicInteger(0);
+  AtomicInteger closeTries = new AtomicInteger(0);
 
   BucketWriter(long rollInterval, long rollSize, long rollCount, long batchSize,
       Context context, String filePath, String fileName, String inUsePrefix,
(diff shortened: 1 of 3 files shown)


szaboferee (changes requested):
Thanks for the contribution!

I have run the tests and `TestHDFSEventSink` with `TestHDFSEventSinkOnMiniCluster` were hanging and resulted in a timeout.

@mcsanady can you have a look at it?

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
```diff
@@ -114,6 +114,7 @@
   // reopened. Not ideal, but avoids internals of owners
   protected boolean closed = false;
   AtomicInteger renameTries = new AtomicInteger(0);
+  AtomicInteger closeTries = new AtomicInteger(0);
```
jojochuang, line 117: This line seems redundant.

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/MockHDFSWriter.java
```diff
@@ -85,6 +101,15 @@ public void sync() throws IOException {
...
   public void close() throws IOException {
     filesClosed++;
+    currentCloseAttempts.incrementAndGet();
+    logger.info("Attempting to close: '" + currentCloseAttempts + "' of '" +
+        numberOfRetriesRequired + "'");
+    if (currentCloseAttempts.get() >= numberOfRetriesRequired || numberOfRetriesRequired == 0) {
+      logger.info("closing file");
+
+    } else {
+      throw new IOException("MockIOException");
```
jojochuang, line 111: This message should log the reason an exception is thrown.

flume-ng-sinks/flume-hdfs-sink/src/test/java/org/apache/flume/sink/hdfs/TestBucketWriter.java
```diff
@@ -450,6 +456,39 @@ public void SequenceFileRenameRetryCoreTest(int numberOfRetriesRequired, boolean
...
+        0, 0, 1, 1, ctx, hdfsPath, hdfsPath, "singleBucket", ".tmp", null, null,
+        null, mockHDFSWriter, timedRollerPool, proxy,
+        new SinkCounter("test-bucket-writer-" + System.currentTimeMillis()), 0, null, null, 30000,
+        Executors.newSingleThreadExecutor(), 1, numberOfRetriesRequired);
+
+    Event event = EventBuilder.withBody("test", Charsets.UTF_8);
+    bucketWriter.append(event);
+    bucketWriter.close(false);
+    TimeUnit.SECONDS.sleep(numberOfRetriesRequired + 2);
+    Assert.assertTrue("Expected " + numberOfRetriesRequired + " " +
```
jojochuang, line 487: This is good, but it would be even better to use Assert.assertEquals()

jojochuang:
@szaboferee 
I ran the two tests you mentioned but they did not fail for me. I'm new to Flume so maybe I missed something. Thanks!

flume-ng-sinks/flume-hdfs-sink/src/main/java/org/apache/flume/sink/hdfs/BucketWriter.java
majorendre: Agree, this is an unused variable.

majorendre:
I can confirm that the org.apache.flume.sink.hdfs.TestHDFSEventSinkOnMiniCluster test hangs.

asfgit:

PR #203, opened 2018-03-19
FLUME-3227 Add Rate Limiter to stresssource

tmgstevens:
This uses a shaded version of the guava rate limiter and includes unit tests.

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/StressSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/StressSource.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import org.apache.flume.source.shaded.guava.RateLimiter;
 import org.apache.flume.ChannelException;
 import org.apache.flume.Context;
 import org.apache.flume.CounterGroup;
@@ -66,6 +67,7 @@ public class StressSource extends AbstractPollableSource implements Configurable
   private Event event;
   private List<Event> eventBatchList;
   private List<Event> eventBatchListToProcess;
+  private RateLimiter limiter;
 
   public StressSource() {
     counterGroup = new CounterGroup();
@@ -88,6 +90,13 @@ public class StressSource extends AbstractPollableSource implements Configurable
     batchSize = context.getInteger("batchSize", 1);
     /* Size of events to be generated. */
     int size = context.getInteger("size", 500);
+
+    int rateLimit = context.getInteger("maxEventsPerSecond", 0);
+    if (rateLimit > 0) {
+      limiter = RateLimiter.create(rateLimit);
+    } else {
+      limiter = null;
+    }
 
     prepEventData(size);
   }
@@ -123,6 +132,9 @@ public class StressSource extends AbstractPollableSource implements Configurable
       lastSent = batchSize;
 
       if (batchSize == 1) {
+        if (limiter != null) {
+          limiter.acquire();
+        }
         getChannelProcessor().processEvent(event);
       } else {
         long eventsLeft = maxTotalEvents - totalEventSent;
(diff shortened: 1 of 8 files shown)


flume-ng-core/src/test/java/org/apache/flume/source/shaded/guava/TestRateLimiter.java
```diff
@@ -0,0 +1,512 @@
+package org.apache.flume.source.shaded.guava;
```
szaboferee, line 1: some tests are commented out, if they are not needed, I believe they could be removed

flume-ng-core/src/main/java/org/apache/flume/source/StressSource.java
```diff
@@ -89,6 +91,13 @@ protected void doConfigure(Context context) throws FlumeException {
     /* Size of events to be generated. */
     int size = context.getInteger("size", 500);
 
+    int rateLimit = context.getInteger("maxEventsPerSecond", 0);
```
szaboferee, line 94: Could you please update the documentation with the new parameter?

flume-ng-core/src/main/java/org/apache/flume/source/StressSource.java
```diff
@@ -123,6 +132,9 @@ protected Status doProcess() throws EventDeliveryException {
       lastSent = batchSize;
 
       if (batchSize == 1) {
+        if (limiter != null) {
```
szaboferee, line 135: Is it intentional to not having a rate limit with batchSize larger than 1?
`limiter.acquire(batchSize)` would be kind of what we need if I'm right

szaboferee (commented):
Thanks for the contribution @tmgstevens !

in general, it looks good with a few comments.
See them in the code.

flume-ng-core/src/test/java/org/apache/flume/source/TestStressSource.java
```diff
@@ -82,6 +82,49 @@ public void testMaxTotalEvents() throws InterruptedException,
     }
     verify(mockProcessor, times(35)).processEvent(getEvent(source));
   }
+  
+  @Test
+  public void testRateLimitedEvents() throws InterruptedException,
```
turcsanyip: This test method covers 2 test cases: rate limited events and non rate limited events.
It would be more straightforward to separate these cases into 2 test methods (1 test case - 1 test method with appropriate names describing the cases).

majorendre:
Hi @tmgstevens, 

Thanks for your contribution. It looks good to me.

Regards,
Endre

turcsanyip (changes requested):
LGTM, except the issues reported by other reviewers and one minor change regarding the test cases / methods.

flume-ng-core/src/test/java/org/apache/flume/source/shaded/guava/TestRateLimiter.java
tmgstevens: Fixed

flume-ng-core/src/main/java/org/apache/flume/source/StressSource.java
tmgstevens: Absolutely, thought I had!!

flume-ng-core/src/main/java/org/apache/flume/source/StressSource.java
tmgstevens: That is a VERY good spot - thanks @szaboferee for being awake when reviewing!!!!!!

It's also slightly trickier than batchSize because it might be the last batch, although perhaps I didn't need to make it quite that complex. Anyhow, done.

flume-ng-core/src/test/java/org/apache/flume/source/TestStressSource.java
tmgstevens: Done. In fact, with the batching changes, now broken into four tests

turcsanyip (approved):
Thanks for the fixes. Everything looks good to me now.

PR #207, opened 2018-05-03
fix bug : Flume TAILDIR source lost last line when file upload from windows

lizhenhuan:
https://issues-test.apache.org/jira/browse/FLUME-3229
Fix this issue .


asfgit:

PR #208, opened 2018-05-30
FLUME-3222 Fix for NoSuchFileException thrown when files are being de…

majorendre:
…leted from the TAILDIR source

We fetch file names from a directory and later we fetch inodes.
If there is a delete between these operations this problem occurs.
Reproduced from unit test.
Added exception handling to handle this case. 
It is enough to ignore the NoSuchFileException and continue.


szaboferee (approved):
@majorendre thanks for the contribution! 
Jenkins build was successful, the test looks ok.
It looks good to me.

turcsanyip:
LGTM, +1

PR #209, opened 2018-05-30
FLUME-3142: Adding HBase2 sink.

turcsanyip:
HBase2Sink is the equivalent of HBaseSink for HBase version 2.
HBaseSink used some API calls which were deprecated in HBase 1.x
and they are not available in HBase 2.x any more.

HBase2Sink has been implemented by copying the existing
flume-ng-hbase-sink module to the new flume-ng-hbase2-sink module,
then adjusting the incompatible API calls to HBase 2.
The package and class names have also been modified to have
the hbase2/HBase2 tag. "Hbase" typos have been fixed too.

The functionality provided by HBase2Sink and the configuration parameters
are the same as in case of HBaseSink (except the hbase2 tag in the sink type
and the package/class names).

HBaseSink has not been modified, so it works with HBase 1.x as before.

### DIFF
--- a/flume-ng-configuration/src/main/java/org/apache/flume/conf/sink/SinkConfiguration.java
+++ b/flume-ng-configuration/src/main/java/org/apache/flume/conf/sink/SinkConfiguration.java
@@ -140,7 +140,7 @@ public class SinkConfiguration extends ComponentConfiguration {
      *
      * @see org.apache.flume.sink.hbase2.HBase2Sink
      */
-    HBASE2("org.apache.flume.sink.hbase2.HBase2Sink"),
+    HBASE2("org.apache.flume.sink.hbase2.HBase2SinkConfiguration"),
 
     /**
      * MorphlineSolr sink
--- a/flume-ng-configuration/src/main/java/org/apache/flume/conf/sink/SinkConfiguration.java
+++ b/flume-ng-configuration/src/main/java/org/apache/flume/conf/sink/SinkConfiguration.java
@@ -136,6 +136,13 @@ public class SinkConfiguration extends ComponentConfiguration {
     ASYNCHBASE("org.apache.flume.sink.hbase.HBaseSinkConfiguration"),
 
     /**
+     * HBase2 sink
+     *
+     * @see org.apache.flume.sink.hbase2.HBase2Sink
+     */
+    HBASE2("org.apache.flume.sink.hbase2.HBase2Sink"),
+
+    /**
      * MorphlineSolr sink
      *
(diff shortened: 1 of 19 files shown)


adenes (approved):
@turcsanyip thank you for the pull request.
The change looks good to me.
The tests in the `flume-ng-hbase2-sink` module pass.
+1

majorendre:
Started testing. Found a problem with packaging: flume-ng-hbase2-sink*.jar is is not added to the distribution package.

majorendre:
New problem found:
Caused by: java.lang.NoSuchMethodException: org.apache.flume.sink.hbase2.HBase2Sink.<init>(java.lang.String)
  at java.lang.Class.getConstructor0(Class.java:3082) ~[?:1.8.0_171]
  at java.lang.Class.getConstructor(Class.java:1825) ~[?:1.8.0_171]
  at org.apache.flume.conf.sink.SinkConfiguration$SinkConfigurationType.getConfiguration(SinkConfiguration.java:190) ~[flume-ng-configuration-1.9.0-SNAPSHOT.jar:1.9.0-SNAPSHOT]

turcsanyip:
> Found a problem with packaging: flume-ng-hbase2-sink*.jar is is not added to the distribution package.

Fixed. Thanks for catching it.

turcsanyip:
> java.lang.NoSuchMethodException: org.apache.flume.sink.hbase2.HBase2Sink.

Fixed. It was a copy-paste error in SinkConfiguration. "HBase2Sink" was configured as the configuration class which led to NoSuchMethodException. 
Defining a non-existing (!) class (like in case of HBase sink and many other sinks) won't cause error and the default SinkConfiguration will be initialized.
However, I don't think this is the best solution for SinkConfiguration and all the enum values should be cleaned up in the future.

majorendre:
My manual tests has passed. Here is what I did:
I deployed the flume dist package on a 4 node hadoop cluster running hbase.
Put an appropriate hbase-site.xml in the flume conf directory.
Created a minimal config for the hbase2 sink
a1.sinks.k1.type = hbase2
a1.sinks.k1.table = foo_table
a1.sinks.k1.columnFamily = bar_cf
a1.sinks.k1.serializer = org.apache.flume.sink.hbase2.SimpleHBase2EventSerializer
Created the necessary table and column family in hbase shell.
Used a spooldir for generating events.
Check the hbase table in the hbase shell. Found all the events there.

szaboferee (approved):
Looks good to me,
tested with the latest hbase2 release from my machine connecting to a remote hbase2 server

PR #212, opened 2018-06-19
FLUME-3246 Validate flume configuration to prevent larger source batc…

majorendre:
…h size than the channel transaction capacity

The loadSources() method seemed like an appropriate place to check this.
Added 2 new interfaces for getting the transaction capacity and the batch size fields. The check is only done for channels that implement the TransactioCapacitySupported interface and sources that implement the BatchSizeSupported  interface.
There is a new unit test case that I used for testing.

TODOs:
Add the BatchSizeSupported interface to all the sources that handle batch size.
Check how this works when reloading configuration.

### DIFF
--- a/flume-ng-core/src/test/java/org/apache/flume/sink/TestRollingFileSink.java
+++ b/flume-ng-core/src/test/java/org/apache/flume/sink/TestRollingFileSink.java
@@ -23,6 +23,7 @@ import org.apache.flume.Channel;
 import org.apache.flume.Context;
 import org.apache.flume.Event;
 import org.apache.flume.EventDeliveryException;
+import org.apache.flume.Transaction;
 import org.apache.flume.channel.MemoryChannel;
 import org.apache.flume.channel.PseudoTxnMemoryChannel;
 import org.apache.flume.conf.Configurables;
@@ -280,6 +281,48 @@ public class TestRollingFileSink {
     File[] files = dir.listFiles();
     for (File file : files) {
       file.delete();
+    }
+  }
+
+  /**
+   * This test is to reproduce batch size and
+   * transaction capacity related configuration
+   * problems
+   */
+  @Test(expected = EventDeliveryException.class)
+  public void testTrans() throws EventDeliveryException {
+
+    Context context = new Context();
+
+    context.put("sink.directory", tmpDir.getPath());
+    context.put("sink.rollInterval", "0");
+    context.put("sink.batchSize", "1000");
+
+    Configurables.configure(sink, context);
+
+    context.put("capacity", "50");
+    context.put("transactionCapacity", "5");
+    Channel channel = new MemoryChannel();
+    Configurables.configure(channel, context);
+
+    sink.setChannel(channel);
+    sink.start();
+
+    try {
+      for (int j = 0; j < 10; j++) {
+        Transaction tx = channel.getTransaction();
+        tx.begin();
+        for (int i = 0; i < 5; i++) {
+          Event event = new SimpleEvent();
+          event.setBody(("Test event " + i).getBytes());
+          channel.put(event);
+        }
+        tx.commit();
+        tx.close();
+      }
+      sink.process();
+    } finally {
+      sink.stop();
     }
   }
 }
--- a/flume-ng-core/src/test/java/org/apache/flume/sink/TestRollingFileSink.java
+++ b/flume-ng-core/src/test/java/org/apache/flume/sink/TestRollingFileSink.java
@@ -290,7 +290,7 @@ public class TestRollingFileSink {
    * problems
    */
   @Test(expected = EventDeliveryException.class)
-  public void testTrans() throws EventDeliveryException {
+  public void testTransCapBatchSizeCompatibility() throws EventDeliveryException {
 
     Context context = new Context();
 
--- a/flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
+++ b/flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
@@ -359,17 +359,13 @@ public abstract class AbstractConfigurationProvider implements ConfigurationProv
   }
 
   private List<Channel> getSourceChannels(Map<String, ChannelComponent> channelComponentMap,
-                                          Source source, Collection<String> channelNames) {
+                  Source source, Collection<String> channelNames) throws InstantiationException {
     List<Channel> sourceChannels = new ArrayList<Channel>();
     for (String chName : channelNames) {
       ChannelComponent channelComponent = channelComponentMap.get(chName);
       if (channelComponent != null) {
-        try {
-          checkSourceChannelCompatibility(source, channelComponent.channel);
-          sourceChannels.add(channelComponent.channel);
-        } catch (InstantiationException e) {
-          LOGGER.error("Source - channel compatibility problem", e);
-        }
+        checkSourceChannelCompatibility(source, channelComponent.channel);
+        sourceChannels.add(channelComponent.channel);
       }
     }
     return sourceChannels;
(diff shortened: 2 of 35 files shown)


flume-ng-core/src/test/java/org/apache/flume/sink/TestRollingFileSink.java
```diff
@@ -282,4 +283,46 @@ private void cleanDirectory(File dir) {
...
     }
   }
+
+  /**
+   * This test is to reproduce batch size and
+   * transaction capacity related configuration
+   * problems
+   */
+  @Test(expected = EventDeliveryException.class)
+  public void testTrans() throws EventDeliveryException {
```
szaboferee: The test method name could tell what is in the comment above

flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
```diff
@@ -278,7 +280,12 @@ private void loadSources(AgentConfiguration agentConf,
           for (String chName : channelNames) {
             ChannelComponent channelComponent = channelComponentMap.get(chName);
             if (channelComponent != null) {
-              sourceChannels.add(channelComponent.channel);
+              try {
```
szaboferee: duplicated code part 1

flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
```diff
@@ -330,7 +337,12 @@ private void loadSources(AgentConfiguration agentConf,
...
```
szaboferee: duplicated code part 2

flume-ng-node/src/test/java/org/apache/flume/node/TestAbstractConfigurationProvider.java
```diff
@@ -185,6 +185,36 @@ public void testSourceAndSinkThrowExceptionDuringConfiguration()
...
+        new MemoryConfigurationProvider(agentName, properties);
+    MaterializedConfiguration config = provider.getConfiguration();
+    Assert.assertTrue(config.getSourceRunners().size() == 1);
+    Assert.assertTrue(config.getChannels().size() == 1);
+    Assert.assertTrue(config.getSinkRunners().size() == 1);
+
+    properties.put(agentName + ".sources.source1.batchSize", "1001");
+    properties.put(agentName + ".sinks.sink1.batch-size", "1001");
+
+    provider = new MemoryConfigurationProvider(agentName, properties);
```
szaboferee: Would it make sense to add assertions to the error messages?
config.getConfigurationErrors()

szaboferee (commented):
Hi @majorendre, thanks for the contribution,
see a few comments in the review. 
I would love to see a test case in flume-ng-tests where a restart/config reload happens.
Otherwise, it looks good to me.

flume-ng-node/src/test/java/org/apache/flume/node/TestAbstractConfigurationProvider.java
```diff
@@ -185,6 +185,36 @@ public void testSourceAndSinkThrowExceptionDuringConfiguration()
     Assert.assertTrue(config.getSinkRunners().size() == 0);
   }
 
+  @Test
+  public void testSinkSourceMismatchDuringConfiguration() throws Exception {
```
turcsanyip, line 189: What happens if there is source - channel mismatch only (or sink - channel only)?
It would be worth splitting the test case to 2 separate cases (or keep this combined case and add 2 more).

flume-ng-node/src/test/java/org/apache/flume/node/TestAbstractConfigurationProvider.java
```diff
@@ -185,6 +185,36 @@ public void testSourceAndSinkThrowExceptionDuringConfiguration()
...
+
+    MemoryConfigurationProvider provider =
+        new MemoryConfigurationProvider(agentName, properties);
+    MaterializedConfiguration config = provider.getConfiguration();
+    Assert.assertTrue(config.getSourceRunners().size() == 1);
+    Assert.assertTrue(config.getChannels().size() == 1);
+    Assert.assertTrue(config.getSinkRunners().size() == 1);
+
+    properties.put(agentName + ".sources.source1.batchSize", "1001");
+    properties.put(agentName + ".sinks.sink1.batch-size", "1001");
```
turcsanyip: Why is it set to 1001?
As far as I understand the point is to exceed the channel capacity, but it is 100 by default, so 101 would also work.
It would be more straightforward to set the channel capacity explicitly in the test instead of using the default one.

flume-ng-sdk/src/main/java/org/apache/flume/api/AbstractRpcClient.java
```diff
@@ -61,4 +64,26 @@ public abstract void appendBatch(List<Event> events)
   protected abstract void configure(Properties properties)
       throws FlumeException;
 
+  public static int parseBatchSize(Properties properties) {
```
turcsanyip: AbstractRpcClient does not seem to be best place for this static method.
_batchSize looks a bit strange in java.
AbstractRpcClient.parseBatchSize(clientProps) call in AbstractRpcSink too.

I would consider to move the method to a utility class (or maybe to RpcClient).

turcsanyip (commented):
Hi @majorendre 

The change looks to me good in general.
Please see some improvement comments below.

flume-ng-sdk/src/main/java/org/apache/flume/api/AbstractRpcClient.java
majorendre: I see your points, but here is some explanation:
AbstractRpcClient.parseBatchSize(clientProps) call exists in AbstractRpcSink because at the time of the config check it does not have a configured client. Otherwise the client handles batch sizes. I will put that in as a comment in the code. 
So the parse belongs to the client as it is their configuration. This is not a generic utility that is why I did not put in a separate class. This part of the flume uses a "batch-size" in the config while other parts use "batchSize".

flume-ng-node/src/test/java/org/apache/flume/node/TestAbstractConfigurationProvider.java
majorendre: agree

flume-ng-core/src/test/java/org/apache/flume/sink/TestRollingFileSink.java
majorendre: agree

flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
majorendre: agree

majorendre:
retest this please

flume-ng-sdk/src/main/java/org/apache/flume/api/AbstractRpcClient.java
turcsanyip: I see, thanks for the explanation.
I did not see that AbstractRpcSink uses different batch size at configuration / processing time (same value, but coming from different places).
The comment wil be fine.

turcsanyip (approved):
Thanks for the changes.
It looks good to me now.

flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
```diff
@@ -364,6 +358,57 @@ private void loadSources(AgentConfiguration agentConf,
...
+  private List<Channel> getSourceChannels(Map<String, ChannelComponent> channelComponentMap,
+                                          Source source, Collection<String> channelNames) {
+    List<Channel> sourceChannels = new ArrayList<Channel>();
+    for (String chName : channelNames) {
+      ChannelComponent channelComponent = channelComponentMap.get(chName);
+      if (channelComponent != null) {
+        try {
+          checkSourceChannelCompatibility(source, channelComponent.channel);
+          sourceChannels.add(channelComponent.channel);
+        } catch (InstantiationException e) {
```
szaboferee: If there is a problem with one of the channels this would just silently remove that one. I think we should let the exception out and make the whole Source removed from the config.

szaboferee (commented):

generally, it looks good to me, just had one comment.

asfgit:

flume-ng-node/src/main/java/org/apache/flume/node/AbstractConfigurationProvider.java
majorendre: I agree.

PR #213, opened 2018-06-29
FLUME-3235.patch Flume TAILDIR lost last line

ffernandez92:
There is an open issue https://issues.apache.org/jira/projects/FLUME/issues/FLUME-3235?filter=allopenissues that reports a problem about avoiding last line when this line does not contain a \n or \r\n.

This commit should fix the problem


asfgit:

PR #216, opened 2018-07-11
startup script supports reading environment variable for some param

zhipcui:


szaboferee:
Hi @zhipcui why do you need this?

These are internal variables and they can be set with different parameters.

What are you trying to achieve?

zhipcui:
All these variables can be set by `flume-env.sh` file on conf dir.
Take `JAVA_OPTS ` as example, if I want to change it, I need `flume-env.sh` file.  
Setting this variables from environment variable would be a more clean way especially run the flume using Kubernetes.

asfgit:

majorendre:
LGTM

turcsanyip:
We can add these environment variables, but I do not think it is really necessary.

Regarding JAVA_OPTS: yes, it is common the pass parameters and system properties to java, but FLUME_JAVA_OPTS can be used for this and can be defined as an environment variable, FLUME_JAVA_OPTS is appended after JAVA_OPTS so it has higher precedence

Regarding classpath: the best practice for using third party libraries is through plugins (by default the plugins.d directory), so adjusting the classpath should be needed in quite rare/special cases only

PR #219, opened 2018-07-27
FLUME-3262: Improve Scribe Source With More Options

JohnZZGithub:
We added following options:
  1. ThriftServerType, user choose one of {THshaServer, TThreadedSelectorServer, TThreadPoolServer} baed on their own traffic pattern. More details on thrift server could be found on Apache Thrift project.
  2. MaxThriftFrameSizeBytes: thrift frame size limit.
  3. MaxReadBufferSize. The buffer size thrift used to handle requests, increase the buffer will benefit throughput under heavy load.

### DIFF
--- a/flume-ng-sources/flume-scribe-source/src/main/java/org/apache/flume/source/scribe/ScribeSource.java
+++ b/flume-ng-sources/flume-scribe-source/src/main/java/org/apache/flume/source/scribe/ScribeSource.java
@@ -19,7 +19,6 @@
 
 package org.apache.flume.source.scribe;
 
-import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -34,13 +33,7 @@ import org.apache.flume.instrumentation.SourceCounter;
 import org.apache.flume.source.AbstractSource;
 import org.apache.flume.source.scribe.Scribe.Iface;
 import org.apache.thrift.TException;
-import org.apache.thrift.protocol.TBinaryProtocol;
-import org.apache.thrift.server.THsHaServer;
-import org.apache.thrift.server.TNonblockingServer;
 import org.apache.thrift.server.TServer;
-import org.apache.thrift.transport.TFramedTransport;
-import org.apache.thrift.transport.TNonblockingServerSocket;
-import org.apache.thrift.transport.TNonblockingServerTransport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,40 +41,23 @@ import org.slf4j.LoggerFactory;
  * Flume should adopt the Scribe entry {@code LogEntry} from existing
  * Scribe system. Mostly, we may receive message from local Scribe and Flume
  * take responsibility of central Scribe.
- *
+ * <p>
  * <p>
  * We use Thrift without deserializing, throughput has 2X increasing
  */
 public class ScribeSource extends AbstractSource implements
-      EventDrivenSource, Configurable {
+    EventDrivenSource, Configurable {
 
   private static final Logger LOG = LoggerFactory.getLogger(ScribeSource.class);
 
-  public static final String SCRIBE_CATEGORY = "category";
-
-  private static final int DEFAULT_PORT = 1499;
-  private static final int DEFAULT_WORKERS = 5;
-  private static final int DEFAULT_MAX_READ_BUFFER_BYTES = 16384000;
-
-  private TServer server;
-  private int port;
-  private int workers;
-  private int maxReadBufferBytes;
-
+  private ScribeSourceConfiguration scribeSourceConfiguration;
   private SourceCounter sourceCounter;
+  private TServer server;
 
   @Override
   public void configure(Context context) {
-    port = context.getInteger("port", DEFAULT_PORT);
-    maxReadBufferBytes = context.getInteger("maxReadBufferBytes", DEFAULT_MAX_READ_BUFFER_BYTES);
-    if(maxReadBufferBytes <= 0){
-      maxReadBufferBytes = DEFAULT_MAX_READ_BUFFER_BYTES;
-    }
-
-    workers = context.getInteger("workerThreads", DEFAULT_WORKERS);
-    if (workers <= 0) {
-      workers = DEFAULT_WORKERS;
-    }
+    scribeSourceConfiguration = new ScribeSourceConfiguration();
+    scribeSourceConfiguration.configure(context);
 
     if (sourceCounter == null) {
       sourceCounter = new SourceCounter(getName());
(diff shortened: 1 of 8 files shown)


asfgit:

flume-ng-sources/flume-scribe-source/src/main/java/org/apache/flume/source/scribe/ScribeSource.java
```diff
@@ -48,71 +48,158 @@
...
-        args.maxWorkerThreads(workers);
-        args.processor(processor);
-        args.transportFactory(new TFramedTransport.Factory(maxReadBufferBytes));
-        args.protocolFactory(new TBinaryProtocol.Factory(false, false));
-        args.maxReadBufferBytes = maxReadBufferBytes;
-
-        server = new THsHaServer(args);
-
+        LOG.info("Initializing thrift server:" + thriftServerType.getValue());
+        switch (thriftServerType) {
```
szaboferee: maybe this could be in a factory to make it readable and more organized.
 
We could expose the ThriftServer configuration parameters in the source params as in the HTTPSource with the FlumeBeanConfigurator
https://github.com/apache/flume/blob/trunk/flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java#L198

szaboferee (changes requested):
@JohnZZGithub thanks for the contribution!

It is a really good feature to have more options.
To make the new feature more accessible for others, could you add the parameters to the documentation with some examples?

It would also be great if you could include some test cases that make sure that the new code works.

We could extract the creation logic into a factory to maintain readability.

flume-ng-sources/flume-scribe-source/src/main/java/org/apache/flume/source/scribe/ScribeSource.java
JohnZZGithub: Sure, will update it later

JohnZZGithub:
The failure seems not to be caused by patch  https://builds.apache.org/job/flume-github-pull-request/org.apache.flume$flume-ng-tests/82/testReport/junit/org.apache.flume.test.agent/TestConfig/testConfigReplacement/

flume-ng-sources/flume-scribe-source/src/main/java/org/apache/flume/source/scribe/ScribeSourceConfiguration.java
```diff
@@ -0,0 +1,120 @@
...
+
+    workerThreadNum = context.getInteger("workerThreads", DEFAULT_WORKERS);
+    if (workerThreadNum <= 0) {
+      workerThreadNum = DEFAULT_WORKERS;
+    }
+    try {
+      thriftServerType = ThriftServerType.getServerType(
+          context.getString("thriftServer", DEFAULT_THRIFT_SERVER_TYPE.getValue()));
+    } catch (IllegalArgumentException ex) {
+      LOG.error("Wrong thrift server type config", ex);
```
majorendre: I would include the bad value in the error message, to help finding the bad config.
Could we throw an exception here instead of the error logging?

JohnZZGithub: @majorendre  Hi, I updated the patch and added an IllegalArgumentException here. However, the caller which is ```  @Override
  public void configure(Context context) ``` can't throw any exception as it overrides an non-exception method.

majorendre: Hi @JohnZZGithub 
Thanks for the quick reply. 
I took a deeper look and I suggest not handling that  IllegalArgumentException at all. You can just let it propagate, Flume will handle it properly. I did test this, to be sure.
See my attached diff. I also added the Apache licence to avoid 'rat' errors.
As @szaboferee mentioned before, please add some test cases.
Regards,
Endre

majorendre:
[scribe-errhand.diff.zip](https://github.com/apache/flume/files/2558653/scribe-errhand.diff.zip)
@JohnZZGithub

turcsanyip:
Hi @JohnZZGithub, thanks for the contribution.
The concept looks fine, but please add tests and documentation as it was mentioned by other reviewers.

I tried to set up Scribe source with different Thrift server types and feed the source with the CLI tool from https://github.com/apache/flume/pull/218.
ThshaServer and TThreadedSelectorServer work fine for me, but I get the following error when I start the source in TThreadPoolServer mode:
`2018-11-09 22:18:58,445 (Thread-2) [WARN - org.apache.thrift.server.TThreadPoolServer.serve(TThreadPoolServer.java:207)] Transport error occurred during acceptance of message.
org.apache.thrift.transport.TTransportException: accept() may not return NULL
	at org.apache.thrift.transport.TServerTransport.accept(TServerTransport.java:62) ~[libthrift-0.9.3.jar:0.9.3]
	at org.apache.thrift.server.TThreadPoolServer.serve(TThreadPoolServer.java:162) [libthrift-0.9.3.jar:0.9.3]
	at org.apache.flume.source.scribe.ScribeSource$Startup.run(ScribeSource.java:78) [flume-scribe-source-1.9.0-SNAPSHOT.jar:1.9.0-SNAPSHOT]
`

JohnZZGithub:
@majorendre  Thanks for the patch, will add more tests.
@turcsanyip  Will test it and get back to you later.

szaboferee:
@JohnZZGithub  could you please rebase your change to the latest trunk? It cannot be applied at the moment.

JohnZZGithub:
@szaboferee  Sure, will probably update the patch today with more tests

JohnZZGithub:
@turcsanyip The failure of TThreadPoolServer is called by wrong socket type (should be blocking server socket), fixed now. Thanks

JohnZZGithub:
Hi, I had updated the patch, please review it when you get a chance, thank you!

szaboferee:
Hi @JohnZZGithub the checks failed. I thought You did not have the time to fix it. I personally start the review after the build is green to save the extra rounds.  Feel free to run a mvn clean verify locally to check if the build is successful before you commit.

JohnZZGithub:
@szaboferee  Ah, sorry, my bad. I had only clicked into the first link and it shows 500 erros. So I thought it was not my patch's problem by mistake. Now I fix it. It's a licences issue.

JohnZZGithub:
The jenkins failure seemed to be caused by following test and not related to my changes.

> [INFO] --- spotbugs-maven-plugin:3.1.7:check (verify.spotbugs) @ flume-kafka-channel ---
> [INFO] BugInstance size is 1
> [INFO] Error size is 16
> [INFO] total 1 violations are found which is set to be acceptable using configured property > maxAllowedViolations :13.
> Below are list of bugs ignored :

> [ERROR]  org.apache.flume.channel.kafka.KafkaChannel$ConsumerAndRecords.getOffsetMapString() makes inefficient use of keySet iterator instead of entrySet iterator > [org.apache.flume.channel.kafka.KafkaChannel$ConsumerAndRecords] At KafkaChannel.java:[line 718] WMI_WRONG_MAP_ITERATOR
> [PMD] No report found for mojo check

JohnZZGithub:
@szaboferee  @majorendre  I updated the diff. Please review it again when you get a chance, thanks.

JohnZZGithub:
@mpercy  Could you please help to review the change? Thanks

PR #222, opened 2018-08-06
FLUME-3050 add counters for error conditions and expose to monitor URL

majorendre:
By introducing error counters it will be easier to monitor problems. Also errors are categorized, hopefully this will help setting up better monitoring solutions.
Concept: an error is when an Exception is thrown or an ERROR level log is written during event processing.
In case of an error at least 1 error counter is increased at least once. (Preferably 1 counter once).
Errors during event processing are counted. Initialization errors are not handled here.
3 types of errors are differentiated.
-Channel read/write errors from the channel when the channel throws a ChannelException.
-Event read/write errors. E.g: A source cannot read an event due to 
-Generic errors - e.g.: TaildirSource cannot write position file.

### DIFF
--- a/flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
+++ b/flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
@@ -75,7 +75,6 @@ import org.slf4j.LoggerFactory;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyListOf;
 import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.when;
 
 public class TestAvroSource {
 
--- a/flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
+++ b/flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
@@ -28,6 +28,7 @@ import java.nio.ByteBuffer;
 import java.security.cert.X509Certificate;
 import java.nio.channels.ServerSocketChannel;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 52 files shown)


turcsanyip (approved):
It looks good to me. The tests passed.
Thanks for the contribution.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
```diff
@@ -64,9 +67,16 @@
...
 import org.junit.Test;
+import org.mockito.Mockito;
+import org.mockito.internal.util.reflection.Whitebox;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyListOf;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.when;
```
szaboferee: unused import

szaboferee (changes requested):

nice change, well tested. 
Could you add the counters to the SequenceGeneratorSource, HiveSink and KafkaSink?
Also it would be great to have a test for the SpoolDirSource as it looks like that it may count some errors both in the Source and the ReliableSpoolingFileEventReader.

asfgit:

majorendre:
Rewrote the SpoolDirectorySource and ReliableSpoolingFileEventReader part.
Only increment the counters in ReliableSpoolingFileEventReader when the problem is handled with logging an error message only. When an exception is thrown, then it will be handled in SpoolDirectorySource, it is much clearer this way. This was one of the first ones I did, and the concept was not ready, yet. Created some Mockito test cases for SpoolDirectorySource changes but I did not create for the ReliableSpoolingFileEventReader part, since it is so simple, and would be quite difficult to test.
I did all the other requested changes, too.

majorendre:
retest this please

PR #223, opened 2018-08-14
FLUME-3268:Introducing micro batch processing to HDFSEventSink

JohnZZGithub:


asfgit:

PR #226, opened 2018-08-28
FLUME-2973 BucketWriter deadlock fix

majorendre:
This PR is based on Yan Jian's fix and his test improvements. Also contains the deadlock reproduction contributed by @adenes.
I have made minimal changes to those contributions.
Denes's test was used for checking the fix.
Yan's fix contains an optimization as it first calls the callback function that removes the BucketWriter from the cache. This is useful, should help to avoid some errors.


turcsanyip (approved):
It looks good to me.

BucketWrites does not hold the two locks simultaneously so deadlock can not happen any more.

Moving the 'remove from cache' step to the beginning of the close() method is also fine / even better as it was earlier, because it minimizes the chance to get the BW instance from cache by another thread while the BW is being closed. If it happens, it is already handled in HDFSEventSink by catching the BucketClosedException.

szaboferee (approved):
Tested it with the attached test application. It work well. Looks good to me.

PR #228, opened 2018-09-25
FLUME-3269: Support JSSE keystore/trustore -D system properties

turcsanyip:
It makes possible to specify global/common SSL keystore parameters (path,
password and type) at Flume agent (process) level for all sources/sinks.
In this way, it is not necessary to define (=copy) the SSL config for each
component in the agent config.

The global SSL parameters can be specified through the standard -D JSSE
system properties or in environment variables.
Component level configuration is still possible.

Priority:
 1. component parameters in agent config
 2. -D system properties
 2. environment variables

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
@@ -41,6 +41,7 @@ import org.apache.flume.instrumentation.SourceCounter;
 import org.apache.flume.source.avro.AvroFlumeEvent;
 import org.apache.flume.source.avro.AvroSourceProtocol;
 import org.apache.flume.source.avro.Status;
+import org.apache.flume.util.SSLUtil;
 import org.jboss.netty.channel.ChannelPipeline;
 import org.jboss.netty.channel.ChannelPipelineFactory;
 import org.jboss.netty.channel.Channels;
@@ -180,9 +181,10 @@ public class AvroSource extends AbstractSource implements EventDrivenSource,
     }
 
     enableSsl = context.getBoolean(SSL_KEY, false);
-    keystore = context.getString(KEYSTORE_KEY);
-    keystorePassword = context.getString(KEYSTORE_PASSWORD_KEY);
-    keystoreType = context.getString(KEYSTORE_TYPE_KEY, "JKS");
+    keystore = context.getString(KEYSTORE_KEY, SSLUtil.getGlobalKeystorePath());
+    keystorePassword = context.getString(KEYSTORE_PASSWORD_KEY,
+        SSLUtil.getGlobalKeystorePassword());
+    keystoreType = context.getString(KEYSTORE_TYPE_KEY, SSLUtil.getGlobalKeystoreType("JKS"));
     String excludeProtocolsStr = context.getString(EXCLUDE_PROTOCOLS);
     if (excludeProtocolsStr == null) {
       excludeProtocols.add("SSLv3");
--- a/flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java
@@ -28,6 +28,7 @@ import org.apache.flume.instrumentation.SourceCounter;
 import org.apache.flume.source.AbstractSource;
 import org.apache.flume.tools.FlumeBeanConfigurator;
 import org.apache.flume.tools.HTTPServerConstraintUtil;
+import org.apache.flume.util.SSLUtil;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.SecureRequestCustomizer;
@@ -130,11 +131,13 @@ public class HTTPSource extends AbstractSource implements
 
       if (sslEnabled) {
         LOG.debug("SSL configuration enabled");
-        keyStorePath = context.getString(HTTPSourceConfigurationConstants.SSL_KEYSTORE);
+        keyStorePath = context.getString(HTTPSourceConfigurationConstants.SSL_KEYSTORE,
+                SSLUtil.getGlobalKeystorePath());
         Preconditions.checkArgument(keyStorePath != null && !keyStorePath.isEmpty(),
                                     "Keystore is required for SSL Conifguration" );
-        keyStorePassword =
-            context.getString(HTTPSourceConfigurationConstants.SSL_KEYSTORE_PASSWORD);
+        keyStorePassword = context.getString(
+                HTTPSourceConfigurationConstants.SSL_KEYSTORE_PASSWORD,
+                SSLUtil.getGlobalKeystorePassword());
         Preconditions.checkArgument(keyStorePassword != null,
             "Keystore password is required for SSL Configuration");
         String excludeProtocolsStr =
--- a/flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
+++ b/flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
@@ -58,6 +58,7 @@ import org.apache.flume.lifecycle.LifecycleState;
 import org.apache.flume.source.avro.AvroFlumeEvent;
 import org.apache.flume.source.avro.AvroSourceProtocol;
 import org.apache.flume.source.avro.Status;
+import org.apache.flume.util.SSLUtil;
 import org.jboss.netty.channel.ChannelPipeline;
 import org.jboss.netty.channel.socket.SocketChannel;
 import org.jboss.netty.channel.socket.nio.NioClientSocketChannelFactory;
@@ -329,7 +330,7 @@ public class TestAvroSource {
   }
 
   @Test
-  public void testSslRequest() throws InterruptedException, IOException {
+  public void testSslRequestWithComponentKeystore() throws InterruptedException, IOException {
 
     Context context = new Context();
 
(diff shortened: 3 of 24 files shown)


flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -998,32 +1110,24 @@ Example for agent named a1:
...
 If the JMS server uses self-signed certificate or its certificate is signed by a non-trusted CA (eg. the company's own
-CA), then a truststore (containing the right certificate) needs to be set up and passed to Flume via the following JSSE
-Java system properties:
-
-.. code-block:: properties
-
-    export JAVA_OPTS="$JAVA_OPTS -Djavax.net.ssl.trustStore=/path/to/truststore.jks"
-    export JAVA_OPTS="$JAVA_OPTS -Djavax.net.ssl.trustStorePassword=password"
+CA), then a truststore (containing the right certificate) needs to be set up and passed to Flume. It can be done via
+the global SSL parameters. For more details about the global SSL setup, see the *SSL/TLS support* section.
```
szaboferee: this and the other references could be a link:
```
see the `SSL/TLS support`_ section
```

szaboferee (approved):
Thanks @turcsanyip the contribution

it looks good to me. there is one minor suggestion in the documentation part

tmgstevens:
@turcsanyip I'm a bit confused by this patch. I've definitely used the -D parameters for specifying the truststore locations and truststore passwords - certainly Kafka (but I also suspect HTTP and Avro) will eventually use the -D locations, as this is standard Java behaviour. I'm not really sure what this is adding. I'm also nervous about clearing the system properties for the law of unintended consequences.

flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
```diff
@@ -41,6 +41,7 @@
 import org.apache.flume.source.avro.AvroFlumeEvent;
 import org.apache.flume.source.avro.AvroSourceProtocol;
 import org.apache.flume.source.avro.Status;
+import org.apache.flume.util.SSLUtil;
```
majorendre, line 44: Unused import

turcsanyip:
@tmgstevens -D parameters does not work for _keystores_ in the current version. If SSL is enabled but the keystore is not defined in the agent config, then we get a config error and the source will not start (HTTPSource, AvroSource, ThriftSource).

Regarding _truststores_, you are right: -D parameters can work for them already in Flume 1.8.
However, it is not really well documented in the User Guide, eg. AvroSink's truststore parameter: _"If not specified, the default Java JSSE certificate authority files (typically “jssecacerts” or “cacerts” in the Oracle JRE) will be used."_
More precisely: the -D parameter will be used, and if it is not defined either, then jssecacerts/cacerts.

The system properties are cleared only in tests (to clean up before the next test case), so it will not affect the production code.

flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
majorendre: sorry, comment went to the wrong file.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
```diff
@@ -58,6 +58,7 @@
 import org.apache.flume.source.avro.AvroFlumeEvent;
 import org.apache.flume.source.avro.AvroSourceProtocol;
 import org.apache.flume.source.avro.Status;
+import org.apache.flume.util.SSLUtil;
```
majorendre: This is unused

flume-ng-doc/sphinx/FlumeUserGuide.rst
turcsanyip: Thanks for the hint. Fixed.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
turcsanyip: Thanks for catching it. Fixed.

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
```diff
@@ -339,7 +339,34 @@ public void testSslRequest() throws InterruptedException, IOException {
...
+
+    doSslRequest();
+  }
+
+  @Test
+  public void testSslRequestWithGlobalKeystore() throws InterruptedException, IOException {
+
+    System.setProperty("javax.net.ssl.keyStore", "src/test/resources/server.p12");
+    System.setProperty("javax.net.ssl.keyStorePassword", "password");
+    System.setProperty("javax.net.ssl.keyStoreType", "PKCS12");
```
tmgstevens, line 353: That's interesting. Does everything support PKCS12 across the board? I guess it does, but we've never tested it.

flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -749,6 +749,102 @@ the selector will attempt to write the events to the optional channels. Any
...
+==================================  ===============================  ==================================
+javax.net.ssl.keyStore              FLUME_SSL_KEYSTORE_PATH          Keystore location
+javax.net.ssl.keyStorePassword      FLUME_SSL_KEYSTORE_PASSWORD      Keystore password
+javax.net.ssl.keyStoreType          FLUME_SSL_KEYSTORE_TYPE          Keystore type (by default JKS)
+javax.net.ssl.trustStore            FLUME_SSL_TRUSTSTORE_PATH        Truststore location
+javax.net.ssl.trustStorePassword    FLUME_SSL_TRUSTSTORE_PASSWORD    Truststore password
+javax.net.ssl.trustStoreType        FLUME_SSL_TRUSTSTORE_TYPE        Truststore type (by default JKS)
+==================================  ===============================  ==================================
+
+The SSL system properties can either be passed on the command line or by setting the ``JAVA_OPTS``
```
tmgstevens, line 804: Is this sh file common to all Flume deployments?

flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java
```diff
@@ -130,11 +131,13 @@ public void configure(Context context) {
...
-        keyStorePath = context.getString(HTTPSourceConfigurationConstants.SSL_KEYSTORE);
+        keyStorePath = context.getString(HTTPSourceConfigurationConstants.SSL_KEYSTORE,
+                SSLUtil.getGlobalKeystorePath());
         Preconditions.checkArgument(keyStorePath != null && !keyStorePath.isEmpty(),
                                     "Keystore is required for SSL Conifguration" );
-        keyStorePassword =
-            context.getString(HTTPSourceConfigurationConstants.SSL_KEYSTORE_PASSWORD);
+        keyStorePassword = context.getString(
+                HTTPSourceConfigurationConstants.SSL_KEYSTORE_PASSWORD,
+                SSLUtil.getGlobalKeystorePassword());
```
tmgstevens, line 140: We should also capture keystore type here, really. That can be added in L207

tmgstevens (commented):
Hi @turcsanyip,
Great patch. Sorry for the confusion - I now get what we're trying to achieve.
Thanks for adding in the blank truststore password tests - that's a useful addition.

Am I right in thinking that we should now support JKS and PKCS12 across all components (HTTP, Avro Source/Sink, Thrift Source/Sink and Kafka Source, Sink and Channel)?

There are no changes to the Kafka components - will these work with the environment variables?

flume-ng-core/src/test/java/org/apache/flume/source/TestAvroSource.java
turcsanyip: I've tested it all components works with PKCS12 (I guess because it is supported at JVM/JDK level).

flume-ng-doc/sphinx/FlumeUserGuide.rst
turcsanyip: flume-env.sh resides in the FLUME_HOME/conf directory and it is passed to the Flume process via the --conf parameter when the agent started.
Typically all processes use this conf directory I think, so the sh is common.
But it is possible to create multiple FLUME_HOME/conf*** directories with different flume-env.sh files and pass them to different processes. However, I don't know if this multiple conf directory pattern is used at all.

flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java
turcsanyip: Good question. HTTPSource does not support the type parameter in its config explicitly, so I did not add it here. We can add it though.

turcsanyip:
PKCS12 is supported by all components.

Regarding Kafka: The SSL handling is quite transparent for Flume, because it happens between the Kafka client and the broker. We just pass the SSL config params (either through Kafka parameters like 'ssl.truststore.location' or -D system properties) to the Kafka client and all other things happen within the client. So we rely on the capabilities of the Kafka client version used by Flume.

Edit:
Sorry, I forgot to answer to the Kafka vs environment variables question: When environment variables are used for setting up SSL, Flume will "convert" them to system properties (as they were passed with -D). Pls. see SSLUtil. initGlobalSSLParameters() which is invoked at startup time. From this point on all components can access the SSL config through the SSL system properties as usual.

majorendre:
Hi @turcsanyip @tmgstevens 
I have just did some checks. I could get some test cases to work with an SSL enabled Kafka with various configuration settings: environment vars, -D params, kafka params. All worked.

flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java
tmgstevens: I think we should add it in for consistency (sorry, I know that's extra work)

tmgstevens:
I'm confused - how can the Kafka Source, Sink and Channel work with environment variables???

tmgstevens:
> Edit:
> Sorry, I forgot to answer to the Kafka vs environment variables question: When environment variables are used for setting up SSL, Flume will "convert" them to system properties (as they were passed with -D). Pls. see SSLUtil. initGlobalSSLParameters() which is invoked at startup time. From this point on all components can access the SSL config through the SSL system properties as usual.

Okay, that figures. I'm happy with that approach. So the only thing missing, I think, is the HTTPSource keystore type.

flume-ng-core/src/main/java/org/apache/flume/source/http/HTTPSource.java
szaboferee: @tmgstevens I have a change on the way where I have added SSL/TLS support for the syslog source. I have made some refactoring that will add this parameter and makes SSL/TLS handling unified. So if it is ok for you I would commit this and add the keystore type with that change.

turcsanyip: I agree with @szaboferee, let's go ahead with committing this PR.

PR #230, opened 2018-10-11
FLUME-3182 add support for SSL/TLS for syslog (tcp) sources

szaboferee:
Adding SSL/TLS support for syslog and multi port syslog sources
This change also contains some refactoring to avoid code duplication in
sources with SSL/TLS support

As SSL/TLS handling is refactored it is convinient to
add the possibility to specify enabled protocol list FLUME-3275 and
add the possibility to specify enabled cipher suite list FLUME-3276

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
@@ -417,11 +417,11 @@ public class AvroSource extends SslContextAwareAbstractSource implements EventDr
     private Supplier<Optional<SSLEngine>> sslEngineSupplier;
 
     public AdvancedChannelPipelineFactory(boolean enableCompression, boolean enableIpFilter,
-        String patternRuleConfigDefinition, Supplier<Optional<SSLEngine>> sslEngineProvider) {
+        String patternRuleConfigDefinition, Supplier<Optional<SSLEngine>> sslEngineSupplier) {
       this.enableCompression = enableCompression;
       this.enableIpFilter = enableIpFilter;
       this.patternRuleConfigDefinition = patternRuleConfigDefinition;
-      this.sslEngineSupplier = sslEngineProvider;
+      this.sslEngineSupplier = sslEngineSupplier;
     }
 
 
--- a/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
@@ -19,7 +19,6 @@
 
 package org.apache.flume.source;
 
-import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.avro.ipc.NettyServer;
 import org.apache.avro.ipc.NettyTransceiver;
@@ -41,7 +40,6 @@ import org.apache.flume.instrumentation.SourceCounter;
 import org.apache.flume.source.avro.AvroFlumeEvent;
 import org.apache.flume.source.avro.AvroSourceProtocol;
 import org.apache.flume.source.avro.Status;
-import org.apache.flume.util.SSLUtil;
 import org.jboss.netty.channel.ChannelPipeline;
 import org.jboss.netty.channel.ChannelPipelineFactory;
 import org.jboss.netty.channel.Channels;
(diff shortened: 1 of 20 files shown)


szaboferee:
some documentation is in progress, I will update the PR with that and the review fixes later

flume-ng-core/src/main/java/org/apache/flume/source/SslContextAwareAbstractSource.java
```diff
@@ -0,0 +1,200 @@
...
+
+  protected SSLParameters getSSLParametersWithEnabledProtocolsAndCipherSuites(
+      SSLParameters sslParameters) {
+    return new SSLParameters(
+        getEnabledCipherSuites(sslParameters.getCipherSuites()),
+        getEnabledProtocols(sslParameters.getProtocols())
+    );
+  }
+
+  private String[] getEnabledProtocols(String[] enabledProtocols) {
```
turcsanyip: The method name is a bit misleading because the input parameter is also 'enabled protocols'.
I would rename it to getFilteredProtocols (+ getFilteredCipherSuites).

flume-ng-core/src/main/java/org/apache/flume/source/SslContextAwareAbstractSource.java
```diff
@@ -0,0 +1,200 @@
...
+    return () -> getSslEngine(useClientMode);
+  }
+
+  protected SSLParameters getSslParametersForContext(SSLContext sslContext) {
+    SSLParameters sslParameters = sslContext.getDefaultSSLParameters();
+
+    return getSSLParametersWithEnabledProtocolsAndCipherSuites(sslParameters);
+  }
+
+  protected SSLParameters getSSLParametersWithEnabledProtocolsAndCipherSuites(
```
turcsanyip: The method name is a bit complicated. On the other hand, the name of the previous method does not describe exactly what it does (following the same logic that should be getSSLParametersWithEnabledProtocolsAndCipherSuitesForContext).
So it is hard to follow what these methods do.

I would simply implement the overloaded version of getEnabledProtocols / getEnabledCipherSuites with SSLParameters input parameter and extract the filtered lists from that object.

flume-ng-sdk/src/main/java/org/apache/flume/api/NettyAvroRpcClient.java
```diff
@@ -96,7 +99,10 @@
   private String truststore;
   private String truststorePassword;
   private String truststoreType;
-  private final List<String> excludeProtocols = new LinkedList<String>();
+  private final Set<String> excludeProtocols = new LinkedHashSet<>(Arrays.asList("SSLv3"));
```
turcsanyip: I think these fields and also some logic (eg. parseList()) could moved to AbstractRPCClient.

flume-ng-core/src/test/java/org/apache/flume/source/TestMultiportSyslogTCPSource.java
```diff
@@ -155,6 +163,103 @@ private static final int getFreePort() throws IOException {
     return portList;
   }
 
+  private List<Integer> testNPortsSSL(MultiportSyslogTCPSource source, Channel channel,
```
turcsanyip: This and testNPorts methods have a lot of common code. As far as I see, testNPortsSSL contains the same code as  testNPorts with some additional logic.
Wouldn't it be possible to consolidate this code duplication?

flume-ng-core/src/main/java/org/apache/flume/source/AvroSource.java
```diff
@@ -455,52 +411,19 @@ private PatternRule generateRule(String patternRuleDefinition) throws FlumeExcep
...
+    public AdvancedChannelPipelineFactory(boolean enableCompression, boolean enableIpFilter,
+        String patternRuleConfigDefinition, Supplier<Optional<SSLEngine>> sslEngineProvider) {
       this.enableCompression = enableCompression;
-      this.enableSsl = enableSsl;
-      this.keystore = keystore;
-      this.keystorePassword = keystorePassword;
-      this.keystoreType = keystoreType;
       this.enableIpFilter = enableIpFilter;
       this.patternRuleConfigDefinition = patternRuleConfigDefinition;
+      this.sslEngineSupplier = sslEngineProvider;
```
turcsanyip: sslEngineProvider => sslEngineSupplier

turcsanyip (changes requested):
Generally looks good to me. Thanks for your contribution.
Please consider my suggestions regarding method names / restructuring and eliminating duplicated code sections (see my comments inline).

majorendre:
It looks good to me apart from the bit of code duplication @turcsanyip mentioned.

majorendre (approved):
It looks good to me apart from the bit of code duplication @turcsanyip mentioned.

turcsanyip (approved):
It looks good to me now.
Thanks for the fixes.

majorendre:
@szaboferee  Thanks for the changes, all looks good.

PR #231, opened 2018-10-29
FLUME-3278 Handling -D keystore parameters in Kafka components

turcsanyip:
Kafka client does not handle -D keystore parameters directly so Flume has to
pass them explicitly in Kafka properties (like ssl.keystore.location, etc).
Also using the same method for the truststore (in order to handle
keystore/truststore in the same way).


szaboferee (approved):
@turcsanyip thanks for your contribution.

It looks good to me. The kafka consumer and producer uses the same SslFactory so the same config is good. The kerberos (SASL) case is handled.

adenes:
👍 🚢 It looks good to me. All the kafka-{sink,channel,source} tests pass.
Thank you @turcsanyip

PR #232, opened 2018-11-07
FLUME-3280 Improve maven build to help code reviews by adding static …

szaboferee:
…code analyzer

Adding spotbugs, pmd to the build.
moving flume-checkstyle to a new build-support module that contains
any future build tool configuration.

Fixed some trivial checkstyle issues.

Updated apache parent version and maven plugin versions.
Added maxAllowedViolations so this commit could contain only the new checks
and not any code change.

Did some cleanup in the pom files.


szaboferee:
I have checked if there is any difference between the jars before and after the change with [jardiff](https://github.com/scala/jardiff) and there was no difference.

majorendre:
@szaboferee This is great. It will help us to improve code quality. 
+1 from me.

turcsanyip:
LGTM +1
Site generation and the reports have been tested / checked.

Great step to try to reach a code quality baseline.
Now we can work on cleaning up the code and reducing the number of the PMD/Spotbugs violations.

PR #233, opened 2018-11-08
FLUME-3282 Use slf4j in every component

szaboferee:
Changed http sink to log with slf4j instead of log4j.
Changed some test to use slf4j as well.


majorendre:
LGTM. +1

turcsanyip:
Thanks for the contribution. Now the logging looks much cleaner.

Wouldn't it be possible to move all log4j exclusions to the dependency management section? In this case it would be handled in one common place, not in the submodules.

turcsanyip (approved):
LGTM +1

Thanks for cleaning up the exclusions / dependency management.

PR #237, opened 2018-11-15
FLUME-2653 Allow hdfs sink inUseSuffix to be empty

majorendre:
This is based on the contributions for FLUME-2653 regarding a new feature for the hdfs sink.
Added a new parameter hdfs.emptyInUseSuffix to allow the output file name to remain unchanged.
See the user guide changes for details.
This is desired feature from the community. 

I added a new junit test case for testing.
Temporarily modified old test cases in my ide to use the new flag, and they passed. I did this just as one of test, to be on the safe side. It is not in this PR.


turcsanyip (approved):
LGTM +1

PR #238, opened 2018-11-15
FLUME-2799 Kafka Source - Add message offset to headers

jbruggem:
It seems when solving https://issues.apache.org/jira/browse/FLUME-2799 , an oversight resulted in the message offset not being added to the header. This change corrects this.

### DIFF
--- a/flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSource.java
+++ b/flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSource.java
@@ -267,6 +267,10 @@ public class KafkaSource extends AbstractPollableSource
           headers.put(KafkaSourceConstants.PARTITION_HEADER,
               String.valueOf(message.partition()));
         }
+        if (!headers.containsKey(OFFSET_HEADER)) {
+          headers.put(OFFSET_HEADER,
+              String.valueOf(message.offset()));
+        }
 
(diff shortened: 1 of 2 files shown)


szaboferee (changes requested):
Thanks for the contribution @jbruggem 

the code looks fine. It just has a few minor issues:
https://builds.apache.org/job/flume-github-pull-request/116/org.apache.flume.flume-ng-sources$flume-kafka-source/pmdResult/new/

jbruggem:
> Thanks for the contribution @jbruggem

thanks for the reply szaboferee. I'm not sure what to do though, because the warnings you reference hereabove is due to the use of a fully qualified name for a constant, which is done throughout the file and in particular in the lines just preceding mine.

What should I do ? Change just my lines and break style with the rest of the file ? Just change the lines around it ? The whole file ?

szaboferee:
There is an ongoing code quality cleanup with the current number of violations allowed for not changing the code at the same time as the quality gate was added. That is why the same violation was not breaking the build, just the new ones you added.

So I would say to just fix the new lines you have added and we will continue the cleanup later.

Thanks

jbruggem:
Done !

szaboferee (approved):

jbruggem:
great ! Do I still have something to do here or will this run its course naturally and by merged ?

turcsanyip (approved):
LGTM +1

Deployed and tested manually.

PR #239, opened 2018-11-21
FLUME-2464 Remove hadoop and hbase profiles.

turcsanyip:
Hadoop 1/2 profiles were obsolete and had not been used for a long time,
so they have been deleted.
HBase profile was always active, so its content has been moved to
top level.
Additional clean-ups: some version declarations moved to the parent pom,
redundant version declarations and exclusions deleted.


szaboferee (approved):
Thanks for the contribution,

It looks good to me.
I have compared the dependency tree before and after. it showed the same artifacts. only the order was different. And some scope was fixed as well

PR #240, opened 2018-11-21
FLUME-3101 Add maxBatchCount config property to Taildir Source.

turcsanyip:
If there are multiple files in the path(s) that need to be tailed and there
is a file written by high frequency, then Taildir can read the batchSize size
events from that file every time. This can lead to an endless loop and Taildir
will only read data from the busy file, while other files will not be
processed.
Another problem is that in this case TaildirSource will be unresponsive to
stop requests too.

This commit handles this situation by introducing a new config property called
maxBatchCount. It controls the number of batches being read consecutively
from the same file. After reading maxBatchCount rounds from a file, Taildir
will switch to another file / will have a break in the processing.

This change is based on hunshenshi's patch.

### DIFF
--- a/flume-ng-doc/sphinx/FlumeUserGuide.rst
+++ b/flume-ng-doc/sphinx/FlumeUserGuide.rst
@@ -1375,7 +1375,10 @@ skipToEnd                           false                          Whether to sk
 idleTimeout                         120000                         Time (ms) to close inactive files. If the closed file is appended new lines to, this source will automatically re-open it.
 writePosInterval                    3000                           Interval time (ms) to write the last position of each file on the position file.
 batchSize                           100                            Max number of lines to read and send to the channel at a time. Using the default is usually fine.
-maxBatchCount                       Long.MAX_VALUE                 Controls the number of batches being read consecutively from the same file. It can be used to prevent reading a file in endless loop when the file is written by high frequency.
+maxBatchCount                       Long.MAX_VALUE                 Controls the number of batches being read consecutively from the same file.
+                                                                   If the source is tailing multiple files and one of them is written at a fast rate,
+                                                                   it can prevent other files to be processed, because the busy file would be read in an endless loop.
+                                                                   In this case lower this value.
 backoffSleepIncrement               1000                           The increment for time delay before reattempting to poll for new data, when the last attempt did not find any new data.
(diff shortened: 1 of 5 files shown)


flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -1401,6 +1402,7 @@ Example for agent named a1:
   a1.sources.r1.headers.f2.headerKey1 = value2
   a1.sources.r1.headers.f2.headerKey2 = value2-2
   a1.sources.r1.fileHeader = true
+  a1.sources.ri.numThreshold = 1000
```
szaboferee: the example is different than the parameter name

flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -1375,6 +1375,7 @@ skipToEnd                           false                          Whether to sk
 idleTimeout                         120000                         Time (ms) to close inactive files. If the closed file is appended new lines to, this source will automatically re-open it.
 writePosInterval                    3000                           Interval time (ms) to write the last position of each file on the position file.
 batchSize                           100                            Max number of lines to read and send to the channel at a time. Using the default is usually fine.
+maxBatchCount                       Long.MAX_VALUE                 Controls the number of batches being read consecutively from the same file. It can be used to prevent reading a file in endless loop when the file is written by high frequency.
```
szaboferee: I would add a little bit more explanation. for example:
If the source is tailing multiple files and one of them is written at a fast rate, it can prevent other files to be read. In this case lower this value.

szaboferee (commented):
Thank you for the contribution @turcsanyip 
It looks good to me, only some minor changes needed.

majorendre:
LGTM, +1.

majorendre (approved):
LGTM, +1.

flume-ng-doc/sphinx/FlumeUserGuide.rst
turcsanyip: Thanks for catching it. Fixed.

flume-ng-doc/sphinx/FlumeUserGuide.rst
turcsanyip: Modified.

PR #241, opened 2018-11-22
FLUME-3294 Fix polling logic in TaildirSource

turcsanyip:
TaildirSource.process() implements the correct polling logic now. It returns
Status.READY / Status.BACKOFF which controls the common backoff sleeping
mechanism implemented in PollableSourceRunner.PollingRunner (instead of
always returning Status.READY and sleeping inside the method which was
an incorrect behaviour).


majorendre (approved):
LGTM

adenes (approved):
Thank you for the patch @turcsanyip.
The new and the existing tests pass, +1 from me.

PR #242, opened 2018-11-22
FLUME-1342 adding jmx metrics tables to docs

majorendre:
This PR adds a few tables to the User Guide that describe the metrics published by sorurces, sinks and channels.
I used simple unix tools to gather the data then I wrote a small utility to convert it to csv.
Then I used an online converter https://www.tablesgenerator.com/ to generate the rst tables and then a little manual editing.
I discovered some rst formatting problems in the FlumeUserGuide.rst, corrected them, too.
It was rather painful process to gather the data and find a decent representation.
So far this PR only contains the end result. I would be happy to share the utilities, just don't know what would be the best way.


adenes:
Thank you @majorendre , this seems to be a really useful addition to our documentation.
LGTM, the `FlumeUserGuide.html` gets rendered correctly including the newly added tables, but I haven't checked their content.

flume-ng-doc/sphinx/FlumeUserGuide.rst
```diff
@@ -4944,6 +4942,176 @@ Monitoring in Flume is still a work in progress. Changes can happen very often.
...
+| KafkaEventGetTimer       |                   |                |           |           |         |        |
++--------------------------+-------------------+----------------+-----------+-----------+---------+--------+
+| OpenConnectionCount      |                   |                |           |           |         |        |
++--------------------------+-------------------+----------------+-----------+-----------+---------+--------+
+
+Sinks 1
+~~~~~~~
+
++------------------------+-------------+------------+--------------+---------------+-------+--------+
+|                        | AbstractRpc | AsyncHBase | BucketWriter | ElasticSearch | HBase | HBase2 |
```
szaboferee: I guess the BucketWriter is the HDFS sink and the AbstractRpc means some other implementation that will be in the metrics. Could you please check it?

szaboferee (commented):
I have one question below, other than that it looks good

flume-ng-doc/sphinx/FlumeUserGuide.rst
majorendre: Thanks for spotting. Corrected them.

PR #243, opened 2018-11-22
FLUME-3243 hdfs.callTimeout deafault increased and deprecated

majorendre:
The default hdfs.callTimeout used by the HDFS sink was too low only 10 seconds that can cause problems on a busy system.
The new default is 30 sec.
I think this parameter should be deprecated and some new more error tolerant solution should be used. To enable the future change I indicated this in the code and in the Users Guide.
Tested only with the unit tests.


szaboferee (approved):

+1

turcsanyip (changes requested):
I think HDFSEventSink.defaultCallTimeout also needs to be set to 30000.

majorendre:
Thanks @turcsanyip. I have no idea, how I could miss that.

turcsanyip (approved):
It looks good to me now. +1

PR #246, opened 2018-11-23
FLUME-2723 batch size trans cap doc update

majorendre:
An update to the configuration section of the user guide.


turcsanyip (approved):
LGTM +1

PR #248, opened 2018-11-26
FLUME-3298 Make hadoop-common optional in hadoop-credential-store-con…

szaboferee:
…fig-filter

hadoop-common should be optional as in the hdfs-sink


majorendre (approved):
LGTM

turcsanyip (approved):
LGTM +1

PR #249, opened 2018-11-26
FLUME-3299 Fix log4j scopes in pom files

szaboferee:
Moving log4j dependencies to test scope.
Adding log4j as dependency to flume-ng-dist to pack it in the binary tarball.


majorendre (approved):
LGTM

turcsanyip (approved):
LGTM

PR #250, opened 2018-11-28
FLUME-3146 Use public API HdfsDataOutputStream#getCurrentBlockReplica…

majorendre:
…tion where applicable

Took over this issue from Wei-Chiu Chuang. Added a few lines to the tests.
All tests pass, the new feature works.


turcsanyip (approved):
LGTM +1

This solution is fine as a fix for the warning log issue, but we could get rid of the whole reflection magic at all. I'll create a separate ticket for it.

PR #257, opened 2018-12-06
Handling doOptions

ritesh-malav:
doOptions handling in flume to not get 403 forbidden for implementing httpservlet


ritesh-malav:
Everytime server makes a connection to flume for putting the message to http handler it will first hit the options url and then get url / post url. As of now this will say 403 forbidden but enabling doOptions would do the trick in this case.

PR #271, opened 2019-01-02
fix bug : Flume Configuration

midKingKing:
I found that i can't load sink properties when i use LOGGER sink because of  here get a 'false' and it result in create a new empty map to cover sinkContextMap


duofu:
在source中通过sigar获取机器信息，但运行时总报：no sigar-amd64-winnt.dll in java.library.path。即使我把sigar-amd64-winnt.dll放到jre/lib目录下，或是flume\lib目录下，或是将sigar-amd64-winnt.dll路径加入到flume-env.ps1中$JAVA_OPTS配置下，均告无效。

PR #272, opened 2019-01-03
FLUME-3271 Update User Guide to follow log4j2 upgrade changes

turcsanyip:
Also add support for configuration via system properties
(similar to the Flume system properties used with Log4j 1.x):
- flume.log.dir
- flume.log.file
- flume.root.logger.level (formerly: flume.root.logger)
- flume.root.logger.appender (formerly: flume.root.logger)

### DIFF
--- a/conf/log4j2.xml
+++ b/conf/log4j2.xml
@@ -18,13 +18,14 @@
 -->
 <Configuration status="ERROR">
   <Properties>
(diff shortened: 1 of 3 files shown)


hanson2021 (commented):
`<Root level="${sys:flume.root.logger.level:-INFO">`
'}' is missed

PR #273, opened 2019-01-04
FLUME-3311 Update User Guide In HDFS Sink

fuhuaiyong:
As the doc describe.
%y	last two digits of year (00..99)
%Y	year (2010)
BUT
there is somthing wrong in 'a1.sinks.k1.hdfs.path = /flume/events/%y-%m-%d/%H%M/%S'

I think the right example is 'a1.sinks.k1.hdfs.path = /flume/events/%Y-%m-%d/%H%M/%S'


szaboferee (approved):
Thanks for the contribution @fuhuaiyong

PR #274, opened 2019-01-30
Flume 1.9

zhangguozhan:
properties: a1.sinks.k1.maxBytesToLog = 160 not effect

SinkConfiguration.
SinkConfigurationType : LOGGER(null), //need add something?
or 
function : SinkConfigurationType.getConfiguration() when  sinkConfigurationName  is null setNotFoundConfigClass()

if (sinkConfigurationName != null) {
          clazz =
              (Class<? extends SinkConfiguration>) Class
                  .forName(sinkConfigurationName);
          instance = clazz.getConstructor(String.class).newInstance(name);
        } else {
//        	return new SinkConfiguration(name);
        	 instance = new SinkConfiguration(name);
        	 instance.setNotFoundConfigClass();  //add this ?
        	 return instance;
        }


szaboferee:
This PR is invalid.

Java-xian-fish:
s

PR #276, opened 2019-01-31
FLUME-3314 Fixed NPE in Kafka source/channel during offset migration

turcsanyip:
Kafka source/channel threw NPE when migrateOffsets() was called
on nonexistent topics.
It has been fixed by adding null check and logging a warning message
(topic not found, skipping offset migration).

After skipping the offset migration, the source/channel works the same way
as the "non offset migration" case:
- starts and prints warning messages about the non-existing topic periodically
- can recover if the topic is created later

### DIFF
--- a/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
+++ b/flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
@@ -342,7 +342,8 @@ public class KafkaChannel extends BasicChannelSemantics {
       // Read the offsets to verify they were committed
       Map<TopicPartition, OffsetAndMetadata> newKafkaOffsets = getKafkaOffsets(consumer);
       logger.debug("Offsets committed: {}", newKafkaOffsets);
-      if (!newKafkaOffsets.keySet().containsAll(zookeeperOffsets.keySet())) {
+      if (newKafkaOffsets == null
+          || !newKafkaOffsets.keySet().containsAll(zookeeperOffsets.keySet())) {
         throw new FlumeException("Offsets could not be committed");
       }
     }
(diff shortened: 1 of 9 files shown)


flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
```diff
@@ -347,13 +351,16 @@ private void migrateOffsets() {
 
   private Map<TopicPartition, OffsetAndMetadata> getKafkaOffsets(
       KafkaConsumer<String, byte[]> client) {
-    Map<TopicPartition, OffsetAndMetadata> offsets = new HashMap<>();
+    Map<TopicPartition, OffsetAndMetadata> offsets = null;
```
szaboferee, line 355: This will cause a NullPointerException on line 345.

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSource.java
```diff
@@ -597,13 +601,16 @@ private void migrateOffsets(String topicStr) {
 
   private Map<TopicPartition, OffsetAndMetadata> getKafkaOffsets(
       KafkaConsumer<String, byte[]> client, String topicStr) {
-    Map<TopicPartition, OffsetAndMetadata> offsets = new HashMap<>();
+    Map<TopicPartition, OffsetAndMetadata> offsets = null;
```
szaboferee, line 605: This will cause a NullPointerException on line 596.

szaboferee (changes requested):
Thanks for the contribution @turcsanyip 
I have found 2 minor issues

flume-ng-channels/flume-kafka-channel/src/main/java/org/apache/flume/channel/kafka/KafkaChannel.java
turcsanyip: It could only happen if the topic gets deleted while the offset migration is running.
Just for sure, I've added the null checks.

szaboferee (approved):

PR #279, opened 2019-03-26
FLUME-3325 fixed TaildirSource oom issue

belveyf:
when tailing a lot of files where some file is writing frequently but the others are idle  can cause the idleInodes List  increasing infinitely, change it to set(list backend) can fix this problem.

### DIFF
--- a/flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/TaildirSource.java
+++ b/flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/TaildirSource.java
@@ -28,6 +28,8 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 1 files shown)


bessbd:
Hi @oldbelvey,

Thank you for this PR! The change looks good, but can you please explain why the change resolves the issue described in the title?

belveyf:
> Hi @oldbelvey,
> 
> Thank you for this PR! The change looks good, but can you please explain why the change resolves the issue described in the title?

Hello  bessbd,  the oom problem is caused by repeatedly add same value into the list. changing list to set  just preventing from adding repeated values.
And it is described more specificly  at  https://issues.apache.org/jira/browse/FLUME-3325

bessbd (approved):
LGTM, +1

Thank you for this change, @oldbelvey

bessbd:
@oldbelvey : can you please include some directions to prove this change resolves the issue?
Ie. how can I reproduce the original issue and see if it happens before applying this change and that it doesn't after applying it?

belveyf:
@bessbd   I don't think it's easy to reproduce.  Tho OOM happens at our production environment, where one flume agent monitored more than fifteen app's log. And as it is described at the JIRA, with one file frequently write while they others are idle, the oom will happen right after the Arraylist boom.

PR #280, opened 2019-04-03
FLUME-3327 add useKafkaHeader option to convert flume headers as kafk…

stakafum:
…a ones transparently in KafkaSource,

KafkaChannel, and KafkaSink

### DIFF
--- a/flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
+++ b/flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
@@ -18,6 +18,7 @@
 
 package org.apache.flume.sink.kafka;
 
+import com.google.common.base.Charsets;
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import org.apache.avro.io.BinaryEncoder;
@@ -437,7 +438,7 @@ public class KafkaSink extends AbstractSink implements Configurable, BatchSizeSu
   private Headers toKafkaHeaders(Map<String, String> headers) {
     Headers kafkaHeaders = new RecordHeaders();
     for (Entry<String, String> header : headers.entrySet()) {
-      kafkaHeaders.add(header.getKey(), header.getValue().getBytes());
+      kafkaHeaders.add(header.getKey(), header.getValue().getBytes(Charsets.UTF_8));
     }
     return kafkaHeaders;
   }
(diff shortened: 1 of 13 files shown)


flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
```diff
@@ -437,7 +438,7 @@ private void setProducerProps(Context context, String bootStrapServers) {
   private Headers toKafkaHeaders(Map<String, String> headers) {
     Headers kafkaHeaders = new RecordHeaders();
     for (Entry<String, String> header : headers.entrySet()) {
-      kafkaHeaders.add(header.getKey(), header.getValue().getBytes());
+      kafkaHeaders.add(header.getKey(), header.getValue().getBytes(Charsets.UTF_8));
```
slem1, line 441: I think expected Charsets should be configurable through flume context parameters

flume-ng-sources/flume-kafka-source/src/main/java/org/apache/flume/source/kafka/KafkaSource.java
```diff
@@ -515,6 +530,14 @@ Properties getConsumerProps() {
     return stringMap;
   }
 
+  private static Map<String, String> toStringMap(Headers kafkaHeaders) {
+    Map<String, String> stringMap = new HashMap<String, String>();
+    for (Header kafkaHeader : kafkaHeaders) {
+      stringMap.put(kafkaHeader.key(), new String(kafkaHeader.value()));
```
slem1, line 536: Should be construct through `public String(byte bytes[], String charsetName)` with a configurable Charset value

slem1:
+1 for this feature ^^

flume-ng-sinks/flume-ng-kafka-sink/src/main/java/org/apache/flume/sink/kafka/KafkaSink.java
tmgstevens: +1

flume-ng-channels/flume-kafka-channel/src/test/java/org/apache/flume/channel/kafka/TestParseAsFlumeEvent.java
```diff
@@ -64,9 +65,12 @@ private void doParseAsFlumeEventFalse(Boolean checkHeaders) throws Exception {
     KafkaProducer<String, byte[]> producer = new KafkaProducer<>(props);
 
     for (int i = 0; i < 50; i++) {
+      RecordHeaders headers = new RecordHeaders();
+      headers.add(KEY_HEADER, (String.valueOf(i) + "-header").getBytes());
```
tmgstevens, line 69: What's the significance of `-header` here? Can we make this configurable, leaving `-header` as default?

kamazee:
Is there a chance for it to get in if all the comments are addressed?

bessbd:
> Is there a chance for it to get in if all the comments are addressed?

I don't have the time and resources to do a proper review here, but when the comments are addressed and @tmgstevens approves the change, then I don't see why not.

tmgstevens:
Yes, I'm happy to push this once the comments have been resolved.

rgoers:
This has conflicts with the patch I just applied for FLUME-3435 which adds support for the timestamp and headers to KafkaSink and KafkaSource. I did not add them to KafkaChannel as I wasn't sure if the semantics should be the same.

PR #281, opened 2019-04-04
FLUME-3328: Fix Deprecated Properties table of HDFS Sink

turcsanyip:


mpercy:
Merged. Thank you for the fix!

PR #282, opened 2019-04-13
FLUME-3330 Number of parameters does not match in logging

xuchuanyin:
number of parameters does not match in logging


mpercy (approved):

mpercy:
Merged. Thank you for the fix!

PR #286, opened 2019-05-23
FLUME-3216.Avoid reading content repeatedly. 

highfei2011:
Reopen this patch.


rgoers:
I am wondering why this was closed? Is the fix not valid. It would have been nice to have a test to validate this.

PR #288, opened 2019-06-29
FLUME-3339:Fix the taildir source read the data duplication.

zhulh200868:
If the file a.log reaches a certain size, it will be renamed a.log.1 and a new file a.log, then a.log and a.log.1 inode are the same, but the file name is different, the default flume taildir source  will think that this is a new file, it will be re-read, that causing data duplication;


zhulh200868:
close

PR #289, opened 2019-06-29
FLUME-3340:The variable 'result' is never used,so delete this variable.

zhulh200868:
The variable 'result' is never used,so delete this variable.


zhulh200868:
close

PR #296, opened 2019-08-23
FLUME-2894 Fix:Flume components should stop in the correct order (gra…

winway:
when use memory channel, the shutdown of sinks will not wait until all events within the channels are consumed, so it cause the data loss.


rgoers:
This PR is not solving the problem reported in FLUME-2894. From what I can tell that was fixed in a different PR.  This PR purports to solve a problem where the memory channel is stopped before all the items in it are removed. The problem is, that is by design. The memory channel is not durable while others like the FileChannel are so it is ok for them to be stopped while still containing events. If this PR were to be merged it would prevent Flume from shutting down until all events can be delivered. In many cases that is not desirable as the Sink consuming events may not be able to consume them due to downstream problems and so Flume would be prevented from shutting down. For this reason I am closing this PR.

PR #297, opened 2019-08-23
FLUME-3320 Fix:Taildir source reset position bug

winway:
Taildir source has chance to reset position to 0 incorrectly, so data while be duplicated. When log rotate happens just in time so that f and tf are not the same file in between code.
```
f (f.length() < tf.getPos()) {
    logger.info("Pos " + tf.getPos() + " is larger than file size! "
        + "Restarting from pos 0, file: " + tf.getPath() + ", inode: " + inode);
    tf.updatePos(tf.getPath(), inode, 0);
}
```


rgoers:
This looks like it is superseded by https://github.com/apache/flume/pull/368.

rgoers:
Superceded by #368

PR #299, opened 2019-09-30
FLUME-3339: Remove xerces and xalan dependencies

tmgstevens:
Xerces and Xalan dependencies reference old versions, but actually aren't used and furthermore are only "provided" scope. Removed these dependencies.


rgoers:
Done in trunk.

PR #304, opened 2019-10-13
Fix broken build

bessbd:
Travis builds have been failing recently with "Expected feature release number in range of 9 to 14, but got: 8".
According to
https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/6
adding "dist: trusty" to .travis.yml should fix the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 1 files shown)


bessbd:
I'm about to merge this.

PR #310, opened 2019-11-19
FLUME-3347:Direct buffer memory causes sink thread to die

lightzhao:
When the java.lang.OutOfMemoryError: Direct buffer memory exception occurs in the hdfsSink thread, the exception is thrown to the SinkRunner thread, but the SinkRunner thread does not catch the Throwable exception, causing the hdfsSink thread to die。

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/SinkRunner.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/SinkRunner.java
@@ -166,6 +166,13 @@ public class SinkRunner implements LifecycleAware {
           } catch (InterruptedException ex) {
             Thread.currentThread().interrupt();
           }
+        } catch (Throwable th){
+          logger.error("sink process failure",th);
+          try {
+            Thread.sleep(maxBackoffSleep);
+          } catch (InterruptedException ex) {
+            Thread.currentThread().interrupt();
+          }
         }
       }
       logger.debug("Polling runner exiting. Metrics:{}", counterGroup);


lightzhao:
exception

lightzhao:
@mpercy Can you help me review?

lightzhao:
@bessbd can you please review?

flume-ng-core/src/main/java/org/apache/flume/SinkRunner.java
```diff
@@ -166,6 +166,13 @@ public void run() {
           } catch (InterruptedException ex) {
             Thread.currentThread().interrupt();
           }
+        } catch (Throwable th){
```
bessbd, line 169: In general, catching `Error` is to be avoided: https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html . Catching the superclass `Throwable` falls under the same category.

Can you think of any alternative ways of improvement?

bessbd:
Thank you for the patch, @lightzhao !

The idea for this change sounds good. Unfortunately, without the right justification, this code probably should not be merged. Can you please think of an alternative improvement?

flume-ng-core/src/main/java/org/apache/flume/SinkRunner.java
mpercy: @lightzhao, OutOfMemoryError means Flume has run out of memory. Can you increase the amount of memory you give to Flume instead?

You also may want to consider more carefully tuning the byteCapacity of your memory channel. See https://flume.apache.org/FlumeUserGuide.html#memory-channel

Either way, I don't think this patch is a good way to work around your OOME problem.

Edit: For some reason I thought you were using the MemoryChannel. Looking again, I guess not. Can you simply give Flume more memory? It seems you are running too close to the memory limit of what you give the JVM via -Xmx.

mpercy (changes requested):

PR #311, opened 2019-11-24
build: fix morphlines build, enable building on JDK 11

mpercy:
This PR includes a morphlines build fix and a couple of changes to enable building on JDK 11, including an upgrade of the Thrift dependency to 0.13.


mpercy:
I know it says the build failed but it's just flaky tests and PMD warnings, so this is a big improvement over straight-up build failure.

bessbd:
LGTM, +1!
Thank you for the patch, @mpercy .

Do we also want to bump java version or just prepare / allow for it?

mpercy:
Hey @bessbd, thanks for the quick review! I'm fine with bumping the Java version but let's do that in a separate PR, since this one is more of a break-fix.

PR #312, opened 2019-12-20
FLUME-3308:TailDirSource will write an empty position json file

felix-thinkingdata:
If an exception occurs while FileWriter writing to the file, whereafter call FileWriter close method ,the file will be cleared. This is FileWriter characteristic.


felix-thinkingdata:
When the number of files exceeds the system limit.

felix-thinkingdata:
Why doesn't it compile

PR #313, opened 2019-12-26
FLUME-3348：TailDirSource will write an empty position when

felix-thinkingdata:
When the number of files exceeds the maximum number of files allowed by the system.

taildirSource throw Exception:

java.io.FileNotFoundException: position/taildir_position.json (Too many open files)
image
the reason:

If an exception occurs while FileWriter writing to the file, whereafter call FileWriter close method ,the file will be cleared. This is FileWriter characteristic.
image


hackallan:
I had the same problem

jvz (approved):

PR #314, opened 2020-01-16
[FLUME-3352]:Unnecessary canary test will block on readonly spooldir …

hackty:
related issue [https://issues.apache.org/jira/projects/FLUME/issues/FLUME-3352](url)

### DIFF
--- a/flume-ng-core/src/main/java/org/apache/flume/client/avro/ReliableSpoolingFileEventReader.java
+++ b/flume-ng-core/src/main/java/org/apache/flume/client/avro/ReliableSpoolingFileEventReader.java
@@ -180,20 +180,22 @@ public class ReliableSpoolingFileEventReader implements ReliableEventReader {
     Preconditions.checkState(spoolDirectory.isDirectory(),
         "Path is not a directory: " + spoolDirectory.getAbsolutePath());
 
-    // Do a canary test to make sure we have access to spooling directory
-    try {
-      File canary = File.createTempFile("flume-spooldir-perm-check-", ".canary",
-          spoolDirectory);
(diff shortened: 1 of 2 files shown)


hackty:
Can you please review? @bessbd @mpercy

hackty:
Could you please review? @bessbd @mpercy

bessbd:
Hi @hackty !

Thank you for this patch! On a first read, it looks good. Can you please add a test to ensure it works correctly and to avoid future changes accidentally removing this fix?

hackty:
@bessbd Thanks for your prompt reply. Two unit tests were added to ensure the correctness on the fix.

The description of two unit test:
**testRenameTrackingPolicyOnReadonlySpoolDirectory**
One supplemented to examine the mechanisation of canary (RenameTrackingPolicy is enabled so the spooldir should have the write access.)
It can pass no matter before or after the fix.

**testTrackerDirTrackingPolicyOnReadonlySpoolDirectory**
One that added to ensure the correctness on currect fix.(TrackerDirTrackingPolicy is enabled and the trackdir is different from spooldir. This time the write permission of spooldir is not necessary to the trackdir and also works well.)
It can fail before the fix and pass after the fix.

hackty:
Could you please review the tests? @bessbd

bessbd:
Thank you for the tests and the descriptions, @hackty .
It looks good to me. I'll leave some time for @mpercy to add his comments.

mpercy:
Patch looks good to me, tests pass. Thanks for the contribution and your patience and persistence in getting this reviewed, @hackty!

mpercy (approved):

PR #321, opened 2020-03-27
ElasticSearch cluster name is optional for REST client

AndreLouisCaron:
When using `.client = rest`, the `.clusterName` argument is never used.
It should be optional in this case.


flume-ng-sinks/flume-ng-elasticsearch-sink/src/main/java/org/apache/flume/sink/elasticsearch/client/ElasticSearchClientFactory.java
```diff
@@ -20,14 +20,16 @@
...
 
 /**
  * Internal ElasticSearch client factory. Responsible for creating instance
  * of ElasticSearch clients.
  */
 public class ElasticSearchClientFactory {
-  public static final String TransportClient = "transport";
-  public static final String RestClient = "rest";
+  public static final String TransportClient = TRANSPORT_CLIENT;
+  public static final String RestClient = REST_CLIENT;
```
AndreLouisCaron, line 32: Note to reviewers: these seem to be used in quite a few places.  Preserving the original names makes the diff smaller.  Let me know if you prefer that I replace the names instead.

bessbd:
Hi @AndreLouisCaron ,
Thank you for this PR! The change looks good to me and it's fairly straightforward. Do you mind adding a test so that this improvement is not accidentally removed? (No doc edit seems necessary as this behaviour is described in the user guide.)

rgoers:

PR #322, opened 2020-03-27
WIP: Honor UUID interceptor in ElasticSearch sink

AndreLouisCaron:
ElasticSearch provides support for deduplication of events using the
`_id` field when inserting documents using the bulk insert facility.

When using the UUID interceptor with `headerName = _id` for this
purpose, the logstash serializer will rename this as `@fields._id`,
which is stored as though it were any other field.

In order to enable deduplication of events in the ElasticSearch sink, we
need to ensure the header containing the document ID is sent as `_id`,
no matter what its original name is.

The dynamic serializer probably does not have the same limitation, but
it currently counts on the UUID interceptor's `headerName` property to
be set to `_id`.  This is not the default and it is unlikely to work in
complex Flume networks that replicate events to multiple sinks.
Therefore, I fixed it to support the same logic as the logstash
serializer.

### DIFF
--- a/flume-ng-sinks/flume-ng-elasticsearch-sink/src/main/java/org/apache/flume/sink/elasticsearch/client/ElasticSearchRestClient.java
+++ b/flume-ng-sinks/flume-ng-elasticsearch-sink/src/main/java/org/apache/flume/sink/elasticsearch/client/ElasticSearchRestClient.java
@@ -20,6 +20,7 @@ package org.apache.flume.sink.elasticsearch.client;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
+import org.apache.commons.lang.StringUtils;
 import org.apache.flume.Context;
 import org.apache.flume.Event;
 import org.apache.flume.EventDeliveryException;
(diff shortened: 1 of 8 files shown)


flume-ng-sinks/flume-ng-elasticsearch-sink/src/main/java/org/apache/flume/sink/elasticsearch/client/ElasticSearchRestClient.java
```diff
@@ -97,11 +106,18 @@ public void addEvent(Event event, IndexNameBuilder indexNameBuilder, String inde
...
+    Map<String, String> headers = event.getHeaders();
     indexParameters.put(INDEX_PARAM, indexNameBuilder.getIndexName(event));
     indexParameters.put(TYPE_PARAM, indexType);
     if (ttlMs > 0) {
       indexParameters.put(TTL_PARAM, Long.toString(ttlMs));
     }
+    if (idHeaderName != null) {
+      String id = headers.get(idHeaderName);
+      if (id != null) {
+        indexParameters.put(ID_PARAM, id);
```
AndreLouisCaron, line 118: Note to reviewers: everything else in this PR exists to support this line.

rgoers:

PR #324, opened 2020-04-06
WIP: Fix PMD violations in Thrift source

AndreLouisCaron:
The affected files seem to be generated by the Thrift compiler.
Assuming there is a more recent version of the Thrift compiler that
emits code that passes the PMD checks, the proper thing to do would be
to upgrade.

I'm not familiar enough with the Java ecosystem and this project's CI to
perform that kind of expiriment at this time.  I'm just trying to fix
the CI build in order to make other changes, so please be lenient :-)

### DIFF
--- a/flume-ng-legacy-sources/flume-thrift-source/src/main/java/com/cloudera/flume/handlers/thrift/EventStatus.java
+++ b/flume-ng-legacy-sources/flume-thrift-source/src/main/java/com/cloudera/flume/handlers/thrift/EventStatus.java
@@ -33,7 +33,7 @@ public enum EventStatus implements org.apache.thrift.TEnum {
 
   private final int value;
 
-  private EventStatus(int value) {
+  EventStatus(int value) {
     this.value = value;
   }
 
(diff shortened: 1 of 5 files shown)


mpercy:
These are generated files and should not be manually edited. If you want to stop the warnings, better to prevent SpotBugs from running on these generated files. I'm not sure how to do it but suspect it should be controllable via the Maven build.

mpercy:
err, PMD

AndreLouisCaron:
Yeah, I figured as much, but:

1) I have no idea what I'm doing;
2) I have very limited time to spend on this; and
3) I wasn't 100% sure this would even fix the build yet.

Now that we see the build passing, I'll mark this as WIP and see if I can exclude the files from the PMD checks.

AndreLouisCaron:
After fiddling around with the PMD plug-in exclusion configuration for over 2 hours, I cannot get it to ignore violations by file and I cannot get it to ignore violations from inner classes (those with a `$` in them).

I did manage to combine a mix of some exclusions and tweaking the threshold, which seems to fix the build.

See #325.

I can ditch this PR if you like the other one better.  Let me know what you think.

AndreLouisCaron:
Closing in favor of #325.

PR #325, opened 2020-04-06
Ignore PMD violations in Thrift source

AndreLouisCaron:
All but one violation were in code generated by the Thrift compiler.
Some of them were excluded specifically using PMD exclusion rules, but
the remaining ones refer to inner classes, which don't seem to be
affected by exclusion rules.


mpercy (approved):

mpercy:
Thanks @AndreLouisCaron for making this change! Looks good to me. Merging.

PR #326, opened 2020-04-14
ref #FLUME-3361 Added support for environment variable in ZooKeeper b…

avinashk07:
Added support for environment variable in ZooKeeper based conf by setting 
-DpropertiesImplementation=org.apache.flume.node.EnvVarResolverProperties . This is consistent with the way environment variables are supported in local file based Flume agent conf.

### DIFF
--- a/flume-ng-node/src/main/java/org/apache/flume/node/AbstractZooKeeperConfigurationProvider.java
+++ b/flume-ng-node/src/main/java/org/apache/flume/node/AbstractZooKeeperConfigurationProvider.java
@@ -74,7 +74,8 @@ public abstract class AbstractZooKeeperConfigurationProvider extends
   protected final String zkConnString;
 
   private static final String DEFAULT_PROPERTIES_IMPLEMENTATION = "java.util.Properties";
-  private static final Logger LOGGER = LoggerFactory.getLogger(AbstractZooKeeperConfigurationProvider.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(
+      AbstractZooKeeperConfigurationProvider.class);
 
   protected AbstractZooKeeperConfigurationProvider(String agentName,
       String zkConnString, String basePath) {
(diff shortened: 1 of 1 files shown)


avinashk07:
Fixed code style issues.

flume-ng-node/src/main/java/org/apache/flume/node/AbstractZooKeeperConfigurationProvider.java
```diff
@@ -90,14 +97,24 @@ protected CuratorFramework createClient() {
...
-    Map<String, String> configMap;
-    if (configData == null || configData.length == 0) {
-      configMap = Collections.emptyMap();
-    } else {
-      String fileContent = new String(configData, Charsets.UTF_8);
-      Properties properties = new Properties();
-      properties.load(new StringReader(fileContent));
-      configMap = toMap(properties);
+    Map<String, String> configMap = Collections.emptyMap(); 
+    if (configData != null && configData.length > 0) {
```
tmgstevens, line 101: The else case previously set configMap = Collections.emptyMap(). We should retain this behaviour.

tmgstevens (commented):
Generally I'm okay with this, other than the point raised. What would the intention be for this though? To support non-properties file configurations? Where does this lead to?

flume-ng-node/src/main/java/org/apache/flume/node/AbstractZooKeeperConfigurationProvider.java
avinashk07: Thanks for taking it up Tristan!
I have initialized configMap to emptyMap (Map<String, String> configMap = Collections.emptyMap();) so that an emptyMap (and not null) is returned even in case of a handled exception.
Tried to explain in detail below. Please let me know if I'm missing anything.

BEFORE:
-------

Map m = null

if config does not exists
	m = empty Map
	return m
else
	m = Map with config value //IO Exception is thrown
	
return m


NOW:
----

Map m = empty Map (so that I don't end up returning null in case of exceptions below)

if config exists
	m = Map with config value
	catch new possible exceptions other than IO
	
return m //so returning empty map like before (retaining the behavior) 
              //if Config does not exist or if an exception occurs that is handled

avinashk07: Hi Tristan, please let me know if you are waiting for anything else from me.

avinashk07:
Hey Tristan,
Please let me know if there anything else I need to fix or explain?

Thanks

PR #333, opened 2020-10-27
fix bug in ElasticSearchSink ,when use IndexnameBuilderContext can't …

zhaifb:
fix bug in ElasticSearchSink ,when use IndexnameBuilderContext can't get the right config.
I think the code is to get the config for IndexNameBuilder  from Context ，but  use  the wrong Context“serializerContext”.


rgoers:
Elasticsearch support has been removed from the main Flume build since it is so out of date. Please work against the flume-search repo.

zhaifb:
ok thanks

PR #334, opened 2020-12-06
Shorten the time that positionFile is empty

zhining-lu:
Shorten the time that positionFile is empty, thereby reducing the risk of positionFile being empty after the program crashes


zhining-lu:
HI, the execution of new FileWriter(file) will clear the positionFile, so I moved this operation to just before writing the positionFile to reduce the probability of positionFile being empty when the program crashes

PR #337, opened 2020-12-25
Enhance taildir source's efficiencies

Chen768959:
Increase the ability to monitor subdirectories

If you need to monitor subdirectories, adding the following parameters into the taildir source configuration file:
filegroupsIncludeChild
filegroupsIncludeChild.<filegroupsIncludeChildName>
The usages of the two parameters are similar to that of filegroups parameter, but they can monitor both existing and newly genereted subdirectories at the same time

This enhancement have no effect to the original taildir source's efficiencies


rgoers:

PR #339, opened 2021-02-20
FLUME-1080：Issue with HDFSEventSink for append support

Taiyi1122:
HDFSEventSink cannot support append file for HDFS, it's a bug because program does not load configuration info.


Taiyi1122 (commented):
Fix bug for FLUME-1080: Issue with HDFSEventSink for append support

busbey:
I think this is intended as a follow on to FLUME-1080?

Please file a new issue in Jira for the problem and please include a test case here that fails without this fix.

PR #351, opened 2022-02-08
FLUME-3415 - Provide Docker image for Flume

tmgstevens:
Draft docker build and integration tests.


busbey:
is there a jira that describes the goals of this PR?

flume-ng-dist/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# under the License.
+#
+
+ARG IMAGE_NAME=adoptopenjdk/openjdk8
+ARG IMAGE_TAG=centos-jre
+
+FROM ${IMAGE_NAME}:${IMAGE_TAG}
+
+ARG ASSEMBLY_VERSION
+ARG MAINTAINER="Apache Flume <dev@flume.apache.org>"
```
busbey, line 25: the MAINTAINER keyword was deprecated a bit ago. I believe the preferred idiomatic way now is via a label:

```
LABEL org.opencontainers.image.authors="dev@flume.apache.org"
```

tmgstevens: Updated

tmgstevens:
Reverting the now duplicated HBase client changes

.github/workflows/build.yml
```diff
@@ -65,14 +65,34 @@ jobs:
...
           -Dsurefire.rerunFailingTestsCount=2
+
+      - name: Deploy Docker (Linux)
+        if: runner.os == 'Linux' && github.ref == 'refs/heads/trunk'
+        timeout-minutes: 90
+        shell: bash
+        run: |
+          ./mvnw deploy -Ddockerfile.username=$${{ secrets.DOCKERHUB_USER }} \
+           -Ddockerfile.password=... $${{ secrets.DOCKERHUB_TOKEN }}
+
```
busbey, line 85: We shouldn't be publishing snapshot builds of trunk to DockerHub. Can we make sure we're only publishing released artifacts?

flume-ng-dist/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+LABEL org.opencontainers.image.authors="${MAINTAINER}"
+LABEL site="https://flume.apache.org"
+
+COPY target/apache-flume-${ASSEMBLY_VERSION}-bin/apache-flume-${ASSEMBLY_VERSION}-bin/ /etc/flume/
```
busbey, line 30: we shouldn't be putting executables into `/etc`. Since we're making a flume user can we put the executables into that user's home directory?

flume-ng-dist/pom.xml
```diff
@@ -91,6 +91,29 @@
           <skip>true</skip>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.spotify</groupId>
+        <artifactId>dockerfile-maven-plugin</artifactId>
+        <version>${dockerfile.version}</version>
+        <!-- Wire up to the default build phases -->
```
busbey, line 98: this plugin is in "public archived" status. we should treat it as EOL and should not rely on it.

busbey: it's over here:
https://github.com/spotify/dockerfile-maven

flume-ng-tests/src/test/java/org/apache/flume/test/agent/TestFileChannelDocker.java
```diff
@@ -0,0 +1,155 @@
...
+   * order.
+   *
+   * @throws Exception
+   */
+  @Test
+  public void testInOut() throws Exception {
+    LOGGER.debug("testInOut() started.");
+
+
+    DockerInstall.getInstance().startAgent("a1", agentProps, tempResources);
```
busbey, line 126: there's a code smell here due to the agent start being in the test while the agent stop is in an after block. either start and stop should both be in the test, or the start should be in a before.

flume-ng-tests/src/test/java/org/apache/flume/test/agent/TestFileChannelDocker.java
```diff
@@ -0,0 +1,155 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(TestFileChannelDocker.class);
+
+  private static final List<File> tempResources = new ArrayList<File>();
+
+  private Properties agentProps;
+  private File sinkOutputDir;
+
+  @Before
+  public void setUp() throws Exception {
+    /* Create 3 temp dirs, each used as value within agentProps */
```
busbey, line 49: you should add a JUnit `Assert.assumeTrue` to skip the test cases when Docker isn't available on the system.

flume-ng-tests/src/test/java/org/apache/flume/test/util/DockerInstall.java
```diff
@@ -0,0 +1,223 @@
...
+    if (containerId == null) {
+      throw new Exception("Process not found");
+    }
+
+    LOGGER.info("Shutting down agent process");
+
+    ImmutableList.Builder<String> builder = new ImmutableList.Builder<>();
+    builder.add("/bin/sh");
+    builder.add("-c");
+    builder.add("'docker kill " +  containerId + "'");
```
busbey, line 76: all of the cases where docker is invoked here should be an environment variable or some such that defaults to "docker" and allows whoever is running the test to specify a fully qualified path to the docker executable.

rgoers:
Now that I look at this again I don't think I really understand how it helps. 

1. It packages all the artifacts whether they are required or not.
2. Unless I am mistaken, it is getting the distribution tar and using the configuration located within that. I fail to see how useful that will be as I would expect most users would have a custom configuration.

To be honest, I added Spring Boot support to simplify a LOT of this. Spring Boot packages up just the dependencies you need, uses the standard application.yml to configure Flume, and allows all the wiring of Sinks, Sources, and Channels to be performed in standard Spring Java config source files. Once packaged with the Spring Boot Maven plugin startup consists of either

1. java -Dspring.profiles.active={envName} -jar my-flume.jar
2../my-flume.jar -Dspring.profiles.active={envName}

busbey:
Container images make deployment to a variety of environments much, much simpler. So I definitely agree with Tristan that we need something like this. For example, it would make it about an order of magnitude easier for me to evaluate release candidates.

Additionally, the project having an opinion on what a reasonable starting point for a container image looks like will make conversation I have about deployment in modern production environments easier.

I think starting from "everything goes in the bucket" to match our historic deployment shape is fine. I agree that eventually we need to get to a more modular approach that provides easy examples for folks building just the parts they need.

There's some tooling around for easier container image building based on spring boot applications, e.g. Jib. Maybe we should take an approach that leverages that?

rgoers:
I certainly don't have an issue against providing a Docker image. I just am unclear as to how it will be useful, but I suppose if it is just used as a starting point it could work out. But the approach you mention sounds much better in the long run. I just don't want to have to support that a) I won't have any use for, and b) we will have to keep compatibility with even though we want to do something much better.

To be clear, I use the FileChannel, which pretty much requires fast disk (i.e. SSDs or equivalent) to perform well. It also requires a dedicated "disk" so that data isn't lost on restart. This doesn't really work well with Docker/Kubernetes so we don't use it for Flume.

tmgstevens:
So a couple of points in here:

> It packages all the artifacts whether they are required or not.

True - packaging is something that I think we need to think about going forwards. For people who want a lightweight deployment, should we offer different profiles, the flipside being that if you're combining two components from different modules (e.g. syslog and kafka, HTTP and HDFS etc) then actually do you ever get the benefit of the modularity, or does it just ramp up your complexity (complexity = adoption blocker in my mind).

> Unless I am mistaken, it is getting the distribution tar and using the configuration located within that. I fail to see how useful that will be as I would expect most users would have a custom configuration.

It does bundle the default conf directory, but it is anticipated that a user would re-map that or pass config in via environment variables (which could then include secrets). Both designed to work in docker and kubernetes. There's an example of doing that here: https://github.com/apache/flume/blob/d2bd7812dbacd86459726c0fd3dc774272ce0222/flume-ng-tests/src/test/java/org/apache/flume/test/util/DockerInstall.java#L137-L153

> I think starting from "everything goes in the bucket" to match our historic deployment shape is fine. I agree that eventually we need to get to a more modular approach that provides easy examples for folks building just the parts they need.

+1

> There's some tooling around for easier container image building based on spring boot applications, e.g. Jib. Maybe we should take an approach that leverages that?

Personally I'd rather not re-write the docker deployment right now given that what's there works pretty well. We could look to move away from the spotify plugin to something else, but I don't want to re-architect the whole packaging of Flume at the moment.

> To be clear, I use the FileChannel, which pretty much requires fast disk (i.e. SSDs or equivalent) to perform well. It also requires a dedicated "disk" so that data isn't lost on restart. This doesn't really work well with Docker/Kubernetes so we don't use it for Flume.

I actually think this would be fine - you can have persistent disks in Kubernetes assigned to pods, same applies to docker where you can mount a volume. Would I necessarily recommend that you re-write your deployment model to use containers? No. But for example, in @busbey 's world where he might be moving from a previously managed deployment to something that needs additional orchestration, using Docker or Kubernetes and deploying agents across many nodes, this could make things much easier to manage and maintain.

PR #352, opened 2022-03-19
FLUME-3408: use jdom2 to avoid xxe issue

pjfanning:
https://issues.apache.org/jira/browse/FLUME-3408

### DIFF
--- a/flume-ng-sinks/flume-ng-morphline-solr-sink/pom.xml
+++ b/flume-ng-sinks/flume-ng-morphline-solr-sink/pom.xml
@@ -96,7 +96,7 @@ limitations under the License.
 
     <dependency>
       <groupId>org.jdom</groupId>
-      <artifactId>jdom</artifactId>
+      <artifactId>jdom2</artifactId>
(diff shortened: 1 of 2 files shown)


busbey:
please run the github actions tests again. You can do this by pushing an empty commit or if you have write access to the repository using the github ui

pjfanning:
@busbey it looks like the version of Tika used by Flume uses an ancient version of Rome and that uses the ancient version of jdom - https://mvnrepository.com/artifact/com.rometools/rome/1.18.0 is a more up to date version of Rome.

Ultimately - Flume NG Morphline Solr Sink will probably need to use a newer version of Solr (so that Tika etc get upgraded)

rgoers:
Flume NG Morphline Solr Sink uses KiteSDK, which is what brings in Tika. Unfortunately, Kite has been abandoned. Extracting the bits needed to upgrade jdom is more than we can do for 1.10.0.

pjfanning:
@rgoers I'll close this and raise a new issue for later

PR #353, opened 2022-03-19
FLUME-3409: httpclient security issue

pjfanning:
https://issues.apache.org/jira/browse/FLUME-3409


busbey:
please run the github actions tests again. You can do this by pushing an empty commit or if you have write access to the repository using the github ui

pjfanning:
Thanks @busbey - I tried the rerun but it still failed.

I'm not really a Flume user - just a random ASF member concerned about projects not using secure dependencies.

PR #367, opened 2022-04-15
[FLUME-3423]:Taildir source will throw unnecessay exception when rest…

jaydyi:
…art agent

### DIFF
--- a/flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/ReliableTaildirEventReader.java
+++ b/flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/ReliableTaildirEventReader.java
@@ -92,8 +92,11 @@ public class ReliableTaildirEventReader implements ReliableEventReader {
     this.fileNameHeader = fileNameHeader;
     updateTailFiles(skipToEnd);
 
-    logger.info("Updating position from position file: " + positionFilePath);
-    loadPositionFile(positionFilePath);
+    File positionFile = new File(positionFilePath);
+    if (positionFile.exists() && positionFile.isFile() && positionFile.length() != 0) {
+      logger.info("Updating position from position file: " + positionFilePath);
+      loadPositionFile(positionFilePath);
+    }
   }
 
   /**


jaydyi:
related issue https://issues.apache.org/jira/browse/FLUME-3423

flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/ReliableTaildirEventReader.java
```diff
@@ -92,8 +92,11 @@ private ReliableTaildirEventReader(Map<String, String> filePaths,
     this.fileNameHeader = fileNameHeader;
     updateTailFiles(skipToEnd);
 
-    logger.info("Updating position from position file: " + positionFilePath);
-    loadPositionFile(positionFilePath);
+    File positionFile = new File(positionFilePath);
+    if (positionFile.exists() && positionFile.isFile() && positionFile.length() != 0) {
+      logger.info("Updating position from position file: " + positionFilePath);
+      loadPositionFile(positionFilePath);
+    }
```
tmgstevens, line 99: Perhaps we should log a warning in the else condition?

jaydyi: I also thought of this at first, but after i read position file operations related logic, i think it's redundant，the reasons are as follows.
When TaildirSource start, operations related to position file include these:

- ReliableTaildirEventReader initialization, `loadPositionFile(positionFilePath)` is executed here.
- PositionWriterRunnable thread launched.
```
positionWriter = Executors.newSingleThreadScheduledExecutor(
        new ThreadFactoryBuilder().setNameFormat("positionWriter").build());
positionWriter.scheduleWithFixedDelay(new PositionWriterRunnable(),
        writePosInitDelay, writePosInterval, TimeUnit.MILLISECONDS);
```
positionWriter's job is to read and write position file, and  every time it is scheduled to execute, if the file does not exist, it will automatically generate an empty file.
So to sum up, in the startup phase, it doesn't matter whether the file exists or not, so even an warning is not needed, because from the user's point of view, it is normal that the file does not exist when I start it.

rgoers:

PR #368, opened 2022-04-27
FLUME-3320 Fix taildir source reset position bug

iijima-satoshi:
https://issues.apache.org/jira/browse/FLUME-3320

### DIFF
--- a/flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/ReliableTaildirEventReader.java
+++ b/flume-ng-sources/flume-taildir-source/src/main/java/org/apache/flume/source/taildir/ReliableTaildirEventReader.java
@@ -39,6 +39,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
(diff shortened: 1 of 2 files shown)


rgoers:

iijima-satoshi:
[#297](https://github.com/apache/flume/pull/297) is duplicated with this and is not sufficient to prevent this problem.
The reason is the following ([my Jira comment](https://issues.apache.org/jira/browse/FLUME-3320?focusedCommentId=17528519&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17528519
)).

> I've met the same problem recently. This is because inode, filesize and lastModifiedTime are retrieved at different times.
> To resolve this completely, it is necessary to retrieve these at the same time. I created a PR for this.

Therefore, if this PR is applied, [#297](https://github.com/apache/flume/pull/297) is unnecessary.

Other 11 PRs are derived from different problems and this PR would not affect the other PRs.

Additionally, I commented some PRs that I can understand.

rgoers:
Is it possible to create a test that fails without this change and succeeds when it is applied? How can I verify this fixes the problem?

PR #369, opened 2022-04-27
FLUME-3060 Fix race condition in flume-taildir-source

iijima-satoshi:
https://issues.apache.org/jira/browse/FLUME-3060


rgoers:
Is it possible to create a test that fails before this fix is applied but succeeds after?

PR #373, opened 2022-06-06
Fix typo in the flume-ng script

bessbd:
There's a typo in the `flume-ng` script: `conf-oassword`. This PR changes that to `conf-password`.

Edit: mind that the typo is local / isolated. It looks fine where it's referenced: https://github.com/apache/flume/blob/55206a891a7159af31243902a09aa05f1f45a138/flume-ng-node/src/main/java/org/apache/flume/node/Application.java#L332


bessbd:
Thank you for the review and merge, @rgoers !

PR #376, opened 2022-06-14
FLUME-3379 Include maven repos needed for transitive dependencies

busbey:
Several dependencies are still old enough that they or their transitive poms refer to HTTP repo urls that get blocked by current versions of maven.

Add HTTPS version where needed for:

* Twitter public repo for ua-parser lib
* Restlet public repo for restlet lib
* Conjars public repo for transitive libs of calcite


tmgstevens:
LGTM

busbey:
would you mind merging @tmgstevens ?

PR #379, opened 2022-08-09
FLUME-3434: TwitterSource fails to cast int to long

tmgstevens:
Adds in counters and improved testing

### DIFF
--- a/flume-ng-sources/flume-twitter-source/pom.xml
+++ b/flume-ng-sources/flume-twitter-source/pom.xml
@@ -50,6 +50,12 @@ limitations under the License.
     </dependency>
 
     <dependency>
+      <groupId>org.easytesting</groupId>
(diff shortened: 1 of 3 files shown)


rgoers:
Does this resolve the problem with the TwitterSource or does the version still need to be upgraded?

tmgstevens:
Yes. The root cause appears to be that Avro is refusing to cast an int to a long, which causes the first serialisation to fail. Then, because of poor exception handling, the writer isn't closed and it fails on all subsequent messages.

I've fixed both of those problems and then enhanced the unit tests to check any errors in the future, although the unit tests only fire if you give them Twitter credentials.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

tmgstevens:
Yes, as in no further changes required and it does fix the problem!

PR #380, opened 2022-08-21
upgrade jar and assembly plugins

hboutemy:
to versions that support Reproducible Builds

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@ limitations under the License.
     <mockito.version>1.9.0</mockito.version>
     <module.name />
     <mvn-antrun-plugin.version>1.8</mvn-antrun-plugin.version>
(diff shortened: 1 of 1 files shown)


tmgstevens:
@hboutemy please can you explain what you mean about reproducibile builds? No issues necessarily, just checking what problem we are solving.

busbey:
Here's the overview. Are these the only plugins we have that weren't already over the minimum version?

https://maven.apache.org/guides/mini/guide-reproducible-builds.html

rgoers:
@hboutemy @tmgstevens Unfortunately I merged this and it broke the Flume release. See https://issues.apache.org/jira/projects/MASSEMBLY/issues/MASSEMBLY-941. I will be reverting the assembly plugin version until this issue is fixed.

PR #381, opened 2022-08-24
Upgrade Protobuf and Guava 

mr1716:
This will update the version of protobuf and Guava to more recent versions


tmgstevens:
@mr1716 I'd avoided doing the protobuf one because I was worried about backwards compatibility - do you think this is a safe change?

mr1716:
@tmgstevens I think it easily could be seen as a safe change because it makes the tool more secure/safe for the end user.

mr1716:
But if protobuf wont be upgraded for backwards compatibility, that makes sense. Will close it then

PR #393, opened 2023-02-01
FLUME-3445 - fix TestTwitterSource.testCarrotDateFormatBug on different locale env

shalk:
fix  FLUME-3445


rgoers:
I see no problems with this.

PR #395, opened 2023-02-13
FLUME-3447. Upgrade commons-text version to fix CVE-2022-42889

nikita15p:
Signed-off-by: nikita15p <37657012+nikita15p@users.noreply.github.com>


rgoers:
Dependabot patch was applied that does the same thing.

PR #397, opened 2023-03-17
[FLUME-3454] fixing exception while starting agent configured to jms source when passwordFile field is missing

nikita15p:
Fixing exception while starting agent configured to jms source when passwordFile field is missing in the flume.conf file


rgoers:
This looks innocent enough. I am just going to merge it.

PR #398, opened 2023-03-18
FLUME-3452. Upgrade libthrift version to fix CVE 2020-1938

xuwenping123:
org.apache.thrift:libthrift:0.14.1 has dependency on tomcat-embed-core : 8.5.46 which is causing CVE 2020-1938.

org.apache.thrift:libthrift:0.14.2 solved the problem for https://mvnrepository.com/artifact/org.apache.thrift/libthrift/0.14.2,

see https://issues.apache.org/jira/browse/FLUME-3452.

### DIFF
--- a/dev-docs/UpdateLicenses.md
+++ b/dev-docs/UpdateLicenses.md
@@ -533,7 +533,7 @@ ALv2. Entry in NOTICE.  Additional entry in NOTICE and LICENSE for
 PureJavaCrc32C; portions 2-Clause BSD.
 
 ```
-   libthrift-0.14.1.jar
+   libthrift-0.14.2.jar
 ```
 
 same as jar-with-dependencies
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@ limitations under the License.
     <solr-global.version>4.3.0</solr-global.version>
     <slf4j.version>1.7.32</slf4j.version>
     <system-rules.version>1.17.0</system-rules.version>
-    <thrift.version>0.14.1</thrift.version>
+    <thrift.version>0.14.2</thrift.version>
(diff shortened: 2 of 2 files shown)


rgoers:
Verified the build is still successful with this change.

rgoers:
The unit tests are failing when run with Java 11 as the truststore is using DSA. They work with Java 8. However, after upgrading from version 265 to 362 of Java 8 the tests fail. So the keystores need to be updated.

xuwenping123:
> The unit tests are failing when run with Java 11 as the truststore is using DSA. They work with Java 8. However, after upgrading from version 265 to 362 of Java 8 the tests fail. So the keystores need to be updated.

I am very sorry for the single test failure caused by my submission.

I continue to focus on the tests involved in this change.

rgoers:
I apologize for making it seem like the unit test failures were due to this change. They were not. I just happened to notice that the CI builds were failing in the Thrift tests. When I used Java 11 on my computer I saw the same failures. Then I noticed the CI build was using a newer version of JDK 8 and when I upgraded to that I saw the same errors. I fixed the problem with FLUME-3460.

xuwenping123:
> I apologize for making it seem like the unit test failures were due to this change. They were not. I just happened to notice that the CI builds were failing in the Thrift tests. When I used Java 11 on my computer I saw the same failures. Then I noticed the CI build was using a newer version of JDK 8 and when I upgraded to that I saw the same errors. I fixed the problem with FLUME-3460.

It's not a problem at all, I wish I could do something with flume.

I did encounter some problems in building the project in the past two days. For details, please refer to this link?
https://github.com/apache/flume/pull/403
or https://issues.apache.org/jira/browse/FLUME-3462

Finally thank you, hope I can do a little work for flume.

PR #405, opened 2023-04-08
[FLUME-3468] fix default value of passwordFile property when userName is set for jms source

nikita15p:
In jms source, when passwordFile property in agent config is not set, currently null value is getting passed which causes exception due to `password = Optional.absent();` in https://github.com/apache/flume/blob/trunk/flume-ng-sources/flume-jms-source/src/main/java/org/apache/flume/source/jms/JMSSource.java#L173. 

This PR, sets default value to empty string in case passwordFile property is not set.


nikita15p:
@rgoers can u please review this

PR #409, opened 2023-05-04
add year 2023 to Notice files

yangjf2019:


nikita15p:
Looks good to me. Cc @busbey @rgoers can u please approve

flume-ng-dist/src/main/appended-resources/META-INF/NOTICE
```diff
@@ -217,8 +217,8 @@ source: https://www.opensource.apple.com/source/configd/configd-453.19/dnsinfo/d
 Scala
 scala-collection-compat
 scala-java8-compat
-Copyright (c) 2002-2022 EPFL
-Copyright (c) 2011-2022 Lightbend, Inc.
+Copyright (c) 2002-2023 EPFL
+Copyright (c) 2011-2023 Lightbend, Inc.
```
busbey: just to confirm, did we update this dependency to a version that changed its copyright year to include 2023?

busbey (changes requested):
you also need to update the year in `flume-ng-clients/flume-ng-log4jappender/src/main/resources-for-jar-with-deps/META-INF/NOTICE`

Can someone confirm that `project.build.outputTimestamp` in `flume-parent/pom` actually gets updated to 2023 for e.g. the generated NOTICE file in jars and the dist archive? Even if it does, do we want to change it to 2023 so that SNAPSHOT builds have a correct year?

yangjf2019:
> flume-ng-clients/flume-ng-log4jappender/src/main/resources-for-jar-with-deps/META-INF/NOTICE

Thanks for your review.I updated.

flume-ng-dist/src/main/appended-resources/META-INF/NOTICE
yangjf2019: This refers to the date of the citation, which I don't think needs to be updated to 2023.

busbey (approved):

rgoers:
The proper way to change this is to change the copyright line from
```
Copyright 2011-2022 The Apache Software Foundation
```
to
```
Copyright 2011-Present The Apache Software Foundation
```
From a legal perspective, copyright notices aren't required to be present in the files. If they are they should contain the year of creation and **may** contain years in which substantial modifications were made. But replacing the ending year with "present" eliminates the need to ever change them again.

NOTICE
```diff
@@ -1,5 +1,5 @@
 Apache Flume
-Copyright 2011-2022 The Apache Software Foundation
+Copyright 2011-2023 The Apache Software Foundation
```
rgoers: Change "2023" to "Present"

flume-ng-clients/flume-ng-log4jappender/src/main/resources-for-jar-with-deps/META-INF/NOTICE
```diff
@@ -1,5 +1,5 @@
 Flume NG Log4j Appender
-Copyright 2009-2022 Apache Software Foundation
+Copyright 2009-2023 Apache Software Foundation
```
rgoers: Same here

rgoers (changes requested):

busbey:
I'm -0 on switching to use "Present". At least my reading of [the published ASF policy on NOTICE files](https://www.apache.org/legal/src-headers.html#notice) requires actual years.

text like "present" is ambiguous because it requires that the consumer of the NOTICE file have additional context about when that file was created / modified / distributed.

yangjf2019:
@busbey  Done, thanks!

PR #418, opened 2023-11-09
[FLUME-3480] remove log4j 1.2.17 due to eol

nikita15p:
ISSUE: log4j 1.x is EOL. It is being used in build-support module and flume-ng-log4jappender. Since the dependency is EOL and vulnerable, this is not allowed to be used in enterprise environment.

SOL: This change will prevent use of log4j 1.x


rgoers:
This PR says log4j 1 is used in the build-support module but the PR doesn't show that.  The Log4j 1 Appender should simmply be removed. Log4j 2 already provides a Flume Appender and now that Flume is part of Logging Services it makes sense for that component to be separated out from the Log4j 2 build.

PR #419, opened 2023-12-21
Use Apache Logging Services reusable workflows

ppkarwasz:
This PR switches Flume to use the reusable workflows from [apache/logging-parent](https://github.com/apache/logging-parent).


rgoers:
I apologize. I never saw this. I have been in the middle of splitting the repo into multiple projects and then got pulled away. I believe the changes you made should apply though so I can give it a try.

ppkarwasz:
The changes might be a little bit out of date. This must have been based on `logging-parent` version `10.0.0` or so.

ppkarwasz:
That is probably the best I can do for now.

The build is failing, but the failures are logged to the Develocity server. You can analyze it at https://ge.apache.org/scans?search.rootProjectNames=Apache%20Flume%20Project&search.timeZoneId=Europe%2FWarsaw#

ppkarwasz:
BTW: I bumped the JDK used by the build (but not the JRE requirement) to 11. Many Maven plugins don't work any more on JDK 8.

PR #427, opened 2026-05-19
Upgrades and simplifies build

ppkarwasz:
The Maven build was stuck at version 2.x of `actions/checkout` and `actions/setup-java`. This change:

- Upgrades the actions to the newest version,
- Adds a test on Windows,
- Simplifies the build by removing the steps that were meant to make more space on the runner. These are not necessarily needed on newer runners.


ppkarwasz:
Merged with #429

PR #429, opened 2026-05-20
Update version to 2.0.0-SNAPSHOT. disable hadoop builds

rgoers:
Update version. Disable the hadoop sinks


ppkarwasz (approved):
LGTM, the only problem, currently is the build.

PR #435, opened 2026-05-28
Use `logging-parent`

ppkarwasz:
Reparent `flume-project` and `flume-parent` to
`org.apache.logging:logging-parent:12.1.1` to align Flume's build with the shared Apache Logging Services parent.

A few checks inherited from `logging-parent` are temporarily disabled (`-Dspotless.skip -Denforcer.skip -Dspotbugs.skip`) and Error Prone is turned off via `-Xlint:all` in the compiler config:

* Spotless / Enforcer: blocked until Spotless is applied across the codebase.
* Error Prone: not previously used by Flume; findings will be fixed in a follow-up.
* SpotBugs: the stricter ruleset surfaces new findings to address later.

A Maven toolchain is also configured temporarily, since several Maven plugins now require Java 17 while some tests still fail on Java 8+.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,14 +56,18 @@ jobs:
         uses: actions/setup-java@v5
         with:
           distribution: ${{ matrix.java-distribution }}
-          java-version: 8
+          java-version: |
+            8
+            17
           cache: maven
 
       - name: Build with Maven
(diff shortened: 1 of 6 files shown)


ppkarwasz:
Closing this: we will reach this status at some point, but using another route.

PR #444, opened 2026-06-10
Move HTTPSource to its own flume-http-source module

ppkarwasz:
Keeping HTTPSource in flume-ng-core forced the heavy Jetty/Gson HTTP stack onto every core consumer. Extract it into a dedicated optional source module under flume-ng-sources, like taildir, so it ships only when needed.

The package (org.apache.flume.source.http) is unchanged, so the SourceType.HTTP reflective mapping and existing "http" configs keep working. Core retains Jetty/Gson for its metrics server and HTTPServerConstraintUtil.

Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


ppkarwasz:
I merged this PR unreviewed, so I can work on #445. The `flume-http-source` will go to the [`loggging-flume-http` repo](https://github.com/apache/logging-flume-http) anyway.

PR #459, opened 2026-06-17
Migrate logging API from SLF4J to Log4j 2.x

ppkarwasz:
Standardize all modules on the Log4j 2.x API so Flume's own logging no longer uses SLF4J. Applied the OpenRewrite Slf4jToLog4j recipe, then normalized the result: drop loggers that were never used, make every logger field private static final named "logger", and use the no-arg LogManager.getLogger() since each one named its own class.

Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/flume-ng-auth/pom.xml
+++ b/flume-ng-auth/pom.xml
@@ -51,7 +51,7 @@
 
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-slf4j-impl</artifactId>
+      <artifactId>log4j-slf4j2-impl</artifactId>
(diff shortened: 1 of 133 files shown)


rgoers (approved):

ppkarwasz:
@rgoers,

I also banned all transitive logging dependencies that are not logging APIs, so Flume libraries don't leak artifacts like `logback-core` or `log4j-slf4j2-impl` to users. It will be up to users to decide whether they want to route Log4j API to SLF4J (`log4j-to-slf4j`) or the other way around (`log4j-slf4j2-impl`).

PR #461, opened 2026-07-02
Centralize third-party dependency versions in the BOM

ppkarwasz:
## Summary

All third-party dependency versions are now declared in a single place, `flume-third-party-dependencies` (the BOM). Previously `flume-parent` and `flume-ng-node` also declared dependency versions directly, some of them duplicating properties already defined in the BOM.

## Changes

- **`flume-third-party/pom.xml`**
  - Absorbs the versions that were living elsewhere: `junit`, `junit-jupiter-*` / `junit-vintage-engine`, `mockito-core`, `fest-reflect`, `system-rules`, `hadoop-minikdc`, `hadoop-mapreduce-client-core`, `apacheds-jdbm1`, `mockserver-netty`, `mockserver-client-java`. The two previously hardcoded versions (`2.0.0-M2`, `3.10.8`) are now properties for consistency.
  - `dependencyManagement` entries are ordered by `artifactId`.
  - Drops the redundant `relativePath` (`../pom.xml` is the Maven default) and clarifies the module description (it is a BOM).
- **`flume-parent/pom.xml`**: `dependencyManagement` now holds only internal Flume modules; removed the leaked entries and their now-unused properties, plus the dead `wiremock.version`. The `curator` / `hadoop` / `system-rules` property duplication with the BOM is gone.
- **`flume-ng-node/pom.xml`**: removed the local `junit-jupiter.version` property and the three inline `<version>` tags; the JUnit 5 deps now inherit the version from the BOM.

## Verification

- `mvn validate` passes across the whole reactor.
- The effective POM for `flume-ng-node` confirms the JUnit 5 artifacts resolve to `5.12.2` via managed dependencies.

Note: `external.protobuf.version` stays in `flume-parent` because it feeds the `protocArtifact` (protoc compiler) config in `flume-file-channel`, which inherits it through the parent chain (BOM properties do not propagate to modules).

### DIFF
--- a/flume-ng-node/pom.xml
+++ b/flume-ng-node/pom.xml
@@ -34,7 +34,6 @@
     <spotbugs.maxAllowedViolations>12</spotbugs.maxAllowedViolations>
     <pmd.maxAllowedViolations>7</pmd.maxAllowedViolations>
     <module.name>org.apache.flume.node</module.name>
-    <junit-jupiter.version>5.12.2</junit-jupiter.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


ppkarwasz:
Closing because `flume-third-party-dependencies` should just contain `runtime` dependencies of the build, not `test` dependencies.

PR #467, opened 2026-07-24
Adapt the build to `logging-parent` conventions

ppkarwasz:
Follow-up to #464: makes the whole reactor build green under the `logging-parent` conventions.

- **Adapt build to `logging-parent` conventions**: `flume-parent` now inherits most build configuration from `logging-parent` and the `apache` parent POM. ZooKeeper is bumped to 3.9.3 to satisfy `requireUpperBoundDeps`.
- **Expand star imports**: required by the formatting rules inherited from `logging-parent`.
- **Fix or suppress Error Prone warnings**:
  - the `GuardedBy` suppressions in `MemoryChannel` and `SpillableMemoryChannel` are intentional, so we add `@SuppressWarnings` annotations,
  - a boxed `Integer` lock in tests have been replaced with an `Object` lock,
  - a never-run test,
  - an ignored return value,
  - and a self-catching assertion.
- **Disable BND baseline for never-released artifacts**: six artifacts have no published version on Maven Central to compare against (verified against `repo1.maven.org`).
- **Configure RAT exclusions via `.rat-excludes`**: the file is read by the Maven build (through a file-activated profile, so repositories without the file are unaffected) and by the ASF Trusted Releases platform RAT check.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,11 @@
+--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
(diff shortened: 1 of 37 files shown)


ppkarwasz:
The use of `.rat-excludes` file instead of adding the exclusion to the Apache RAT configuration is compatible with the changes proposed in apache/logging-parent#502.

rgoers (approved):

PR #470, opened 2026-07-28
Build with the `logging-parent` reusable workflows

ppkarwasz:
Replaces the hand-rolled `build.yml` with `build.yaml`, delegating to the `logging-parent` reusable workflows the way `logging-log4j2` does.

### What changes

* `build` now calls `build-reusable.yaml`, replacing the local checkout / `setup-java` / `mvnw verify` steps.
* New `deploy-snapshot` (on `trunk`) and `deploy-release` (on `release/*`) jobs.
* New `verify-reproducibility-snapshot` and `verify-reproducibility-release` jobs.

### Notes for reviewers

* **The build now runs on three operating systems.** `build-reusable.yaml` hardcodes `[macos-latest, ubuntu-latest, windows-latest]` and exposes no input to narrow it, whereas the old workflow was Ubuntu-only. The comment blaming protobuf 2.x for the disabled macOS build was stale: Flume is on protobuf 4.35.0, which ships an `osx-aarch_64` `protoc`. Windows is untested, and several modules depend on Hadoop, which commonly wants `winutils.exe` and `HADOOP_HOME`.


rgoers (approved):

ppkarwasz:
The last commit disables tests that depend on a UNIX file system: TaildirSource tracks files by inode (`unix:ino`), so `TestTaildirSource` and `TestTaildirEventReader` are now skipped when the default file system does not provide a `unix` attribute view (notably on Windows). The Taildir section of the user guide now documents this requirement.

PR #503, opened 2026-09-17
Moved HTTP Source to logging-flume-http

rgoers:


ppkarwasz (approved):

flume-ng-configuration/src/main/java/org/apache/flume/conf/source/SourceType.java
```diff
@@ -89,7 +89,7 @@ public enum SourceType implements ComponentWithClassName {
      *
      * @see org.apache.flume.source.http.HTTPSource
      */
-    HTTP("org.apache.flume.source.http.HTTPSource"),
+    HTTP("org.apache.flume.http.source.HTTPSource"),
```
ppkarwasz, line 92: BTW: we should probably replace this hardcoded list with a `ServiceLoader` lookup.

